PR #11, opened 2018-10-03
Add Codecov.io support

RobAtticus:


RobAtticus:
Not an amazing start, but a good baseline for us to work from :)

PR #18, opened 2018-10-12
Refresh the tsbs_generate_data/devops pkg + add tests

RobAtticus:
Many unnecessary fields were exported, things that should have
been constants or global (unexported) vars were not, and each
subsystem lacked any testing. Most of these things on a subsystem
basis have been corrected. A few files could use some more de-
exporting of vars and other cleanups.

Additionally, there were some minor sanity-related errors in the
memory and disk subsystems. It was relatively easy for the basic
checks like total = free + used to fail, and some values were put
under the wrong labels in the resulting Points. The simpler ones
were corrected.

### DIFF
--- a/cmd/tsbs_generate_data/devops/mem.go
+++ b/cmd/tsbs_generate_data/devops/mem.go
@@ -1,7 +1,6 @@
 package devops
 
 import (
-	"math"
 	"math/rand"
 	"time"
 
@@ -12,11 +11,10 @@ import (
 var (
 	labelMem = []byte("mem") // heap optimization
 
-	// Choices for modeling a host's memory capacity.
-	MemoryMaxBytesChoices = []int64{8 << 30, 12 << 30, 16 << 30}
+	// memoryTotalChoices are the choices for modeling a host's total memory capacity.
+	memoryTotalChoices = []int64{8 << 30, 12 << 30, 16 << 30}
 
-	// Field keys for 'mem' points.
-	MemoryFieldKeys = [][]byte{
+	memoryFieldKeys = [][]byte{
 		[]byte("total"),
 		[]byte("available"),
 		[]byte("used"),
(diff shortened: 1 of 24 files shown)


RobAtticus:
@LeeHampton if you want to specifically look at the changes to `disk`/`disk_test` and `mem`/`mem_test` to make sure I didn't introduce any other logic bugs, it'd be appreciated. Most of the other subsystems are straightforward.

cmd/tsbs_generate_data/devops/mem_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	want = int64(math.Floor(m.distributions[2].Get()))
+	if got := p.GetFieldValue(bufferedKey); got != want {
+		t.Errorf("buffered data out of sync with distribution: got %d want %d", got, want)
+	}
+	total := p.GetFieldValue(totalKey).(int64)
+	used := p.GetFieldValue(usedKey).(int64)
+	available := p.GetFieldValue(availableKey).(int64)
+	if total-int64(used) != int64(available) {
+		t.Errorf("memeory semantics do not make sense: %d - %d != %d", total, used, available)
```
LeeHampton: *memory

cmd/tsbs_generate_data/devops/mem.go
```diff
@@ -60,17 +59,21 @@ func (m *MemMeasurement) ToPoint(p *serialize.Point) {
 	p.SetTimestamp(&m.timestamp)
 
 	total := m.bytesTotal
-	used := m.distributions[0].Get()
-	cached := m.distributions[1].Get()
-	buffered := m.distributions[2].Get()
+	used := int64(math.Floor(m.distributions[0].Get()))
+	cached := int64(math.Floor(m.distributions[1].Get()))
```
LeeHampton: Do we need `math.Floor()` here? Won't casting a float to an int always do the same thing? Or is this one of those weird things where floating point arithmetic sometimes gets weird?

svenklemm: afaik its only different for negative values

Issue #6, opened 2018-08-17
Add ClickHouse to benchmark


jack-pappas:
[ClickHouse](https://clickhouse.yandex/) is another time-series database used for metrics storage and analysis on a large scale. For example, CloudFlare is storing ~6PiB of analytics data in a ClickHouse cluster:
https://blog.cloudflare.com/http-analytics-for-6m-requests-per-second-using-clickhouse/

ClickHouse also has a benchmarks/comparisons page showing performance with a number of different queries on a test dataset -- perhaps those can be included in your benchmark suite as well:
https://clickhouse.yandex/benchmark.html

sunsingerus:
I am trying to implement ClickHouse benchmark functionality and have couple of questions regarding the code.
1. Can you, please, explain, what is the main idea behind writing generated data in text format that would be next parsed by loader? For example, for timescaledb, would it not be simpler, to generate ready-to-use `INSERT` data statements, which loader would just execute?
2. scripts/*.sh files are implemented with different code conventions. For example: `generate_data.sh` and `generate_queries.sh` are using 
```bash
dataDir=${dataDir:-/tmp}
```
convention, while `load_timescaledb.sh`
```bash
DATA_FILE_NAME=${DATA_FILE_NAME:-timescaledb-data.gz}
```
Which is the preferable one?

sunsingerus:
@RobAtticus thank you for explanations.

1. Regarding intermediate CSV-like format. Point taken, it is a nice idea to implement and test different  target DB schemes out of the single test data set. Yes, I've seen that with TimescaleDB - two options available - either to keep the whole record in one JSON field, or use relational approach. This seems applicable with ClickHouse as well - tags can keept tags either in _classic_ tables (and use `JOIN`) or in _dictionaries_ [nice feature of ClickHouse](https://clickhouse.yandex/docs/en/operations/table_engines/dictionary/). 

I am still in the early research/design stage, mainly looking over the code, so more questions may come later.

2. Regarding shell scripts code conventions - that's not a problem at all, would make `load_clickhouse.sh` and was wondering, what style is the preferred one.

sunsingerus:
Hi, I am still pushing ClickHouse test little-by-little, and it is moving. So, may be next week, we can take a look on a newcomer. Pull Request would be not a small one, may take some time to review and adopt!

RobAtticus:
[...]

PR #25, opened 2018-10-23
Fileio

sunsingerus:
Provide possibility to specify input/output filename for plain text files for all tools:
1. data generator
2. query generator
3. data loader
4. query runner

New option added: `-file /path/to/file` (short form) or `--file=/path/to/file` (long form)

More discussion in [this issue](https://github.com/timescale/tsbs/issues/6#issuecomment-431955616)

Use-cases:
```
./tsbs_generate_data -format timescaledb -use-case cpu-only -scale-var 10 -seed 123 -file /tmp/bulk_data/timescaledb
./tsbs_generate_queries -format timescaledb -use-case cpu-only -scale-var 10 -seed 123 -query-type lastpoint -file /tmp/bulk_data/query_timescaledb
./tsbs_load_timescaledb --postgres="sslmode=disable port=5433" --db-name=benchmark --host=127.0.0.1 --user=postgres --workers=1 --file=/tmp/bulk_data/timescaledb_data
./tsbs_run_queries_timescaledb --postgres="sslmode=disable port=5433" --db-name=benchmark --hosts=127.0.0.1 --user=postgres --workers=1 --file=/tmp/bulk_data/timescaledb_query
```

PS. Sorry for play with PR(s) I am still learning on how to perform inter-repo communication in github (that's why extra commits)

### DIFF
--- a/cmd/tsbs_generate_data/main.go
+++ b/cmd/tsbs_generate_data/main.go
@@ -47,7 +47,7 @@ const (
 	errInvalidGroupsFmt = "incorrect interleaved groups configuration: id %d >= total groups %d"
 	errInvalidFormatFmt = "invalid format specifier: %v (valid choices: %v)"
 
-	inputBufSize = 4 << 20
+	defaultWriteSize  = 4 << 20 // 4 MB
 )
 
 // semi-constants
@@ -85,6 +85,7 @@ var (
 	interleavedGenerationGroups  uint
 
 	logInterval time.Duration
+	fileName    string
 )
 
 func parseTimeFromString(s string) time.Time {
@@ -134,6 +135,22 @@ func postFlagParse(flags parseableFlagVars) {
 	timestampEnd = parseTimeFromString(flags.timestampEndStr)
 }
 
+// GetBufferedWriter returns the buffered Writer that should be used for generated output
+func GetBufferedWriter(fileName string) *bufio.Writer {
+	// Prepare output file/STDOUT
+	if len(fileName) > 0 {
+		// Write output to file
+		file, err := os.Create(fileName)
+		if err != nil {
+			fatal("cannot open file for write %s: %v", fileName, err)
+		}
+		return bufio.NewWriterSize(file, defaultWriteSize)
+	}
+
+	// Write output to STDOUT
+	return bufio.NewWriterSize(os.Stdout, defaultWriteSize)
+}
+
 // Parse args:
 func init() {
 	pfv := parseableFlagVars{}
@@ -156,6 +173,8 @@ func init() {
 	flag.StringVar(&profileFile, "profile-file", "", "File to which to write go profiling data")
 
 	flag.DurationVar(&logInterval, "log-interval", 10*time.Second, "Duration between host data points")
+	flag.StringVar(&fileName, "file", "", "File name to write generated data to")
+
 	flag.Parse()
 
 	postFlagParse(pfv)
@@ -174,11 +193,13 @@ func main() {
 	}
 
 	rand.Seed(seed)
-	out := bufio.NewWriterSize(os.Stdout, inputBufSize)
+
+	// Get output writer
+	out := GetBufferedWriter(fileName)
 	defer func() {
 		err := out.Flush()
 		if err != nil {
-			log.Fatal(err.Error())
+			fatal(err.Error())
 		}
 	}()
 
--- a/cmd/tsbs_generate_queries/main.go
+++ b/cmd/tsbs_generate_queries/main.go
@@ -41,12 +41,17 @@ var useCaseMatrix = map[string]map[string]utils.QueryFillerMaker{
 	},
 }
 
+const defaultWriteSize  = 4 << 20 // 4 MB
+
 // Program option vars:
 var (
+	fatal     = log.Fatalf
+
 	generator utils.DevopsGenerator
 	filler    utils.QueryFiller
 
 	queryCount int
+	fileName   string
 
 	seed  int64
 	debug int
@@ -75,6 +80,22 @@ func getGenerator(format string, start, end time.Time, scale int) utils.DevopsGe
 	}
 
 	panic(fmt.Sprintf("no devops generator specified for format '%s'", format))
+}
+
+// GetBufferedWriter returns the buffered Writer that should be used for generated output
+func GetBufferedWriter(fileName string) *bufio.Writer {
+	// Prepare output file/STDOUT
+	if len(fileName) > 0 {
+		// Write output to file
+		file, err := os.Create(fileName)
+		if err != nil {
+			fatal("cannot open file for write %s: %v", fileName, err)
+		}
+		return bufio.NewWriterSize(file, defaultWriteSize)
+	}
+
+	// Write output to STDOUT
+	return bufio.NewWriterSize(os.Stdout, defaultWriteSize)
 }
 
 // Parse args:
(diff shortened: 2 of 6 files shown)


cmd/tsbs_generate_data/main.go
```diff
@@ -174,14 +178,31 @@ func main() {
...
+
+	// Prepare output file/STDOUT
+	var out  *bufio.Writer
+	var file io.Writer
+	var err  error
+	if len(fileName) > 0 {
+		// Write output to file
+		file, err = os.Create(fileName)
+		if err != nil {
+			panic("cannot open file " + fileName)
```
RobAtticus: We are using `fatal()` here so it can 1) be tested and 2) has a prettier output than panic. Also it would be helpful if the error message included the err, so maybe `fatal("cannot open file %s: %v", fileName, err)`

cmd/tsbs_generate_data/main.go
```diff
@@ -174,14 +178,31 @@ func main() {
 	}
 
 	rand.Seed(seed)
-	out := bufio.NewWriterSize(os.Stdout, inputBufSize)
+
+	// Prepare output file/STDOUT
+	var out  *bufio.Writer
+	var file io.Writer
+	var err  error
```
RobAtticus: only `var file io.Writer` is needed here, you only use `err` in the `if` branch so you can use `:=` there, and `out` is assigned afterward so you can use `:=` there too

cmd/tsbs_generate_data/main.go
```diff
@@ -48,6 +48,7 @@ const (
 	errInvalidFormatFmt = "invalid format specifier: %v (valid choices: %v)"
 
 	inputBufSize = 4 << 20
+	defaultWriteSize  = 4 << 20 // 4 MB
```
RobAtticus: these seem to be duplicates, I think you can remove my `inputBufSize` since your name is actually more accurate

cmd/tsbs_generate_queries/main.go
```diff
@@ -41,12 +42,17 @@ var useCaseMatrix = map[string]map[string]utils.QueryFillerMaker{
 	},
 }
 
+const(
+	defaultWriteSize  = 4 << 20 // 4 MB
```
RobAtticus: since its the only `const`, you can just do the one-liner form

load/loader_test.go
```diff
@@ -102,13 +102,13 @@ func TestGetBufferedReader(t *testing.T) {
...
 	// TODO Filename not yet supported
-	r.filename = "foo"
+	r.fileName = "foo"
 	br = r.GetBufferedReader()
 	if br != nil {
 		t.Errorf("filename returned a non-nil buffered reader")
 	}
 	// Should give a non-nil bufio.Reader now
-	r.filename = ""
+	r.fileName = "/dev/null"
```
RobAtticus: I think this case is supposed to test whether reading from STDIN works, so I don't think you should change the case here.

load/loader_test.go
```diff
@@ -102,13 +102,13 @@ func TestGetBufferedReader(t *testing.T) {
 		t.Errorf("initial buffered reader is non-nil")
 	}
 	// TODO Filename not yet supported
-	r.filename = "foo"
+	r.fileName = "foo"
 	br = r.GetBufferedReader()
 	if br != nil {
```
RobAtticus: shouldn't this return a non-nil value now that you'd added support for it? So shouldn't the test be failing?

cmd/tsbs_generate_queries/main.go
```diff
@@ -159,9 +167,28 @@ func main() {
 	// Set up bookkeeping:
 	stats := make(map[string]int64)
 
+	// Prepare output file/STDOUT
+	var out  *bufio.Writer
```
RobAtticus: same things mentioned above.

cmd/tsbs_generate_data/main.go
sunsingerus: Let's completely rewrite this section. I'll extract in to separate function like GetBufferedWriter() as an analogue to GetBufferedReader(). Later we can extract those functions one layer higher into `common`  or `util`

cmd/tsbs_generate_data/main.go
sunsingerus: The same as above - extract to separate function

cmd/tsbs_generate_data/main.go
sunsingerus: Ok, thanks - removed

cmd/tsbs_generate_queries/main.go
sunsingerus: Agree, made it as a one-liner

load/loader_test.go
sunsingerus: Reader can not return reasonable value in here, because there is no "foo" file to read. Anyway, I've remade this part with `fatal()` check - as I've seen in other test case.

RobAtticus: Can we switch `"foo"` to `"/dev/null"` then? Or make a temp file and use it (just have to remember to remove the file too)?

sunsingerus:
Changes made and pushed. Failed test  now passes as
```
=== RUN   TestGetBufferedReader
--- PASS: TestGetBufferedReader (0.01s)
```
Not sure why it fails with DB tests now, need to take a look.

cmd/tsbs_generate_data/main.go
RobAtticus: Good idea -- I do have plans to refactor the two generate binaries into a common lib like `load` and `query`

load/loader_test.go
sunsingerus: Let's not `change` "foo: to "/dev/null", but `add` "/dev/null" as a 3-rd test. So we'll have
1. test for STDIN - should be positive
2. test for non-existing file "foo" - should be negative
3. test for existing file "/dev/null" - should be positive
Let me check

load/loader_test.go
sunsingerus: Agree, my mistake. Sure we need to test for STDIN

load/loader_test.go
sunsingerus: Added `/dev/null`, so there are 3 test cases now:
1. `STDIN`
2. non-existing file `foo`
3. existing file `/dev/null`

Travis CI reported
```
=== RUN   TestGetBufferedReader
--- PASS: TestGetBufferedReader (0.01s)
```

RobAtticus:
@sunsingerus The test is failing because it expects it to `panic` which doesn't happen since you changed some of the `panic`s in `load` pkg to `fatal`s. For now maybe change those back to `panic`s and open an issue for me to make them consistent in the future. Don't want this PR to balloon up too much in scope.

sunsingerus:
@RobAtticus yes, made some arrangements in `panic` vs `fatal` in `load` pkg. Checks passed, it's better to return to `fatal`s later

load/loader.go
```diff
@@ -120,9 +125,16 @@ func (l *BenchmarkRunner) RunBenchmark(b Benchmark, workQueues uint) {
 // GetBufferedReader returns the buffered Reader that should be used by the loader
 func (l *BenchmarkRunner) GetBufferedReader() *bufio.Reader {
 	if l.br == nil {
-		if len(l.filename) > 0 {
-			l.br = nil // TODO - Support reading from files
+		if len(l.fileName) > 0 {
+			// Read from specified file
+			file, err := os.Open(l.fileName)
+			if err != nil {
+				fatal("cannot open file" + l.fileName)
```
RobAtticus: this should be the same as the other:
`fatal("cannot open file %s: %v", l.fileName, err)`

query/benchmarker.go
```diff
@@ -84,6 +90,24 @@ type Processor interface {
...
 }
 
+// GetBufferedReader returns the buffered Reader that should be used by the loader
+func (b *BenchmarkRunner) GetBufferedReader() *bufio.Reader {
+	if b.br == nil {
+		if len(b.fileName) > 0 {
+			// Read from specified file
+			file, err := os.Open(b.fileName)
+			if err != nil {
+				panic("cannot open file" + b.fileName)
```
RobAtticus: same as above

RobAtticus:
Looks great, just two nits left

load/loader.go
sunsingerus: ok, done

query/benchmarker.go
sunsingerus: yes, done

sunsingerus:
@RobAtticus yes, indeed, those error messages can be made more detailed. Changes pushed. Those `getBufferedWriter()` functions looks like good candidacies to be extracted into one common pkg one level up.

RobAtticus:
For sure, will think on how to arrange the packages in the near future

RobAtticus:
No worries, you would just squash/rebase as normal, then use `git push -f` (the force flag) to overwrite the branch on github. So your github branch `fileio` will be overwritten with the new, squashed commit.

sunsingerus:
Yes, thank you for the advice, it worked well - squashed and force pushed. Thanks again! One commit with all changes looks like ready!

PR #26, opened 2018-10-24
Add a feature ClickHouse database support

sunsingerus:
ClickHouse can be tested along with all other DBses.
ClickHouse implementation is based on TimescaleDB.


sunsingerus:
Implementation of [this issue](https://github.com/timescale/tsbs/issues/6)

RobAtticus:
Thanks for this! May be a few days before we get to this, but nice to see!

One note, for the commit msgs what I meant was "Add [a feature]" where [a feature] is replaced with what you are implementing. In this case "Add support for Clickhouse benchmarking"

sunsingerus:
> ... a few days before we get to this...
Yes, sure, no problem at all

> ... "Add [a feature]" where [a feature] is replaced with what you are implementing. In this case "Add support for Clickhouse benchmarking"
Understood, sorry, missed this idea. Hope to be more precise next time

sunsingerus:
Fixed commit message, looks better now

cmd/tsbs_generate_queries/databases/clickhouse/devops.go
```diff
@@ -0,0 +1,390 @@
...
+// 	AND time < '$HOUR_END'
+// GROUP BY minute
+// ORDER BY minute ASC
+func (d *Devops) GroupByTime(qi query.Query, nHosts, numMetrics int, timeRange time.Duration) {
+	interval      := d.Interval.RandWindow(timeRange)
+	metrics       := devops.GetCPUMetricsSlice(numMetrics)
+	selectClauses := d.getSelectClausesAggMetrics("max", metrics)
+
+	sql := fmt.Sprintf(`
+		SELECT
```
RobAtticus: This format seems very weird and hard to parse.

cmd/tsbs_generate_queries/databases/timescaledb/devops.go
```diff
@@ -70,19 +73,32 @@ const goTimeFmt = "2006-01-02 15:04:05.999999 -0700"
...
 	interval := d.Interval.RandWindow(timeRange)
 	metrics := devops.GetCPUMetricsSlice(numMetrics)
 	selectClauses := d.getSelectClausesAggMetrics("max", metrics)
 
-	sql := fmt.Sprintf(`SELECT time_bucket('1 minute', time) AS minute,
-    %s
-    FROM cpu
-    WHERE %s AND time >= '%s' AND time < '%s'
-    GROUP BY minute ORDER BY minute ASC`,
+	sql := fmt.Sprintf(`
```
RobAtticus: I would prefer these formats _not_ to change since I find this new way much more difficult to read.

cmd/tsbs_generate_queries/uses/devops/common.go
```diff
@@ -22,22 +22,22 @@ const (
 	// DoubleGroupByDuration is the how big the time range for DoubleGroupBy query is
 	DoubleGroupByDuration = 12 * time.Hour
 	// HighCPUDuration is the how big the time range for HighCPU query is
-	HighCPUDuration = 12 * time.Hour
+	HighCPUDuration       = 12 * time.Hour
```
RobAtticus: was this run through `gofmt`? This is not the way my `gofmt` treats these

cmd/tsbs_generate_queries/uses/devops/common.go
```diff
@@ -157,41 +161,45 @@ func GetMaxAllLabel(dbName string, nHosts int) string {
...
 	return hostnames
 }
 
-// getRandomSubsetPerm returns a subset of nItems of a permutation of numbers
-// from 0 to scale. This is an alternative to rand.Perm and then taking a
-// sub-slice, which used up a lot more memory and slowed down query generation
-// significantly. The subset of the permutation should have no duplicates.
-func getRandomSubsetPerm(scale, nItems int) []int {
-	if nItems > scale {
+// getRandomSubsetPermutation returns a subset of numItems of a permutation of numbers from 0 to totalNumbersScale.
```
RobAtticus: I don't think this func needs a name change

cmd/tsbs_generate_queries/uses/devops/common.go
```diff
@@ -157,41 +161,45 @@ func GetMaxAllLabel(dbName string, nHosts int) string {
 	return fmt.Sprintf("%s max of all CPU metrics, random %4d hosts, random %s by 1h", dbName, nHosts, MaxAllDuration)
 }
 
-func getRandomHosts(scale, nHosts int) []string {
-	if nHosts < 1 {
-		fatal("number of hosts cannot be < 1; got %d", nHosts)
+// getRandomHosts returns a subset of numHosts hostnames of a permutation of hostnames,
+// numbered from 0 to totalHostsScale.
+// Ex.: host_12, host_7, host_22 for numHosts=3 and totlaHostsScale=30
+func getRandomHosts(numHosts int, totalHostsScale int) []string {
```
RobAtticus: Changing the order of the arguments in this file worries me -- are you sure you got all the places where it may have changed?

cmd/tsbs_generate_queries/uses/devops/groupby_orderby_limit.go
```diff
@@ -12,7 +12,9 @@ type GroupByOrderByLimit struct {
 
 // NewGroupByOrderByLimit returns a new GroupByOrderByLimit for given paremeters
 func NewGroupByOrderByLimit(core utils.DevopsGenerator) utils.QueryFiller {
-	return &GroupByOrderByLimit{core}
+	return &GroupByOrderByLimit{
```
RobAtticus: I'm not a huge fan of these changes, seems unnecessary?

cmd/tsbs_load_timescaledb/main.go
```diff
@@ -98,7 +98,9 @@ func init() {
 type benchmark struct{}
 
 func (b *benchmark) GetPointDecoder(br *bufio.Reader) load.PointDecoder {
-	return &decoder{scanner: bufio.NewScanner(br)}
+	return &decoder{
```
RobAtticus: These changes seem unnecessary. Let's revert to keep the scope of this PR smaller.

load/creator.go
```diff
@@ -5,10 +5,13 @@ package load
 type DBCreator interface {
 	// Init should set up any connection or other setup for talking to the DB, but should NOT create any databases
 	Init()
+
```
RobAtticus: Not sure this is necessary?

load/duplex_channel.go
```diff
@@ -9,11 +9,11 @@ type duplexChannel struct {
 	toScanner chan bool
 }
 
-// newDuplexChannel returns a duplexChannel whose buffer sizes are equal to queue
-func newDuplexChannel(queue int) *duplexChannel {
+// createDuplexChannel returns a duplexChannel with specified buffer sizes
```
RobAtticus: Name change seems unnecessary to me; I prefer `new` to `create` here (`NewFoo`/`newFoo` seems pretty standard golang convention)

load/loader.go
```diff
@@ -13,13 +13,13 @@ import (
 )
 
 const (
-	// DefaultBatchSize is the default size of batches to be inserted
+	// Default size of batches to be inserted
 	defaultBatchSize = 10000
 	defaultReadSize  = 4 << 20 // 4 MB
 
-	// WorkerPerQueue is the value to have each worker have its own queue of batches
+	// Each worker has its own queue of batches
```
RobAtticus: Golang doc comments should begin with the name of the item I believe, so these comments are not in line with that.

load/scan.go
```diff
@@ -5,10 +5,10 @@ import (
 	"reflect"
 )
 
-// ackAndMaybeSend adjust the outstanding batches count and potentially sends
-// another batch to the worker via ch. If unsent is non-empty, send. Returns
-// the updated state of unsent
-func ackAndMaybeSend(ch *duplexChannel, count *int, unsent []Batch) []Batch {
+// sendOutstandingBatchToWorker adjust the outstanding batches count and potentially sends
```
RobAtticus: Not a fan of this name change. Less accurate than before.

load/scan.go
```diff
@@ -24,27 +24,28 @@ func ackAndMaybeSend(ch *duplexChannel, count *int, unsent []Batch) []Batch {
 // sendOrQueueBatch attempts to send a Batch of data on a duplexChannel; if it would block
 // or there is other work to be sent first, the Batch is stored on a queue. The count of
 // outstanding work is adjusted upwards
-func sendOrQueueBatch(ch *duplexChannel, count *int, batch Batch, unsent []Batch) []Batch {
+func sendOrQueueBatch(ch *duplexChannel, count *int, batch Batch, outstandingBatches []Batch) []Batch {
```
RobAtticus: don't think this parameter name change is improvement.

In general we try to keep the names somewhat short

load/scan.go
```diff
@@ -99,28 +102,35 @@ func scanWithIndexer(channels []*duplexChannel, batchSize uint, limit uint64, br
...
 	// We use Select via reflection to either select an acknowledged channel so
 	// that we can potentially send another batch, or if none are ready to continue
 	// on scanning. However, when we reach a limit of outstanding batches, we also
 	// want to block until one worker is done, so as not to starve the workers.
-	// Using an array with Select via reflection gives us this flexibility (i.e., we can
-	// either pass the whole array of cases, or the array less the last item).
-	cases := make([]reflect.SelectCase, numChannels+1)
+	// Using an array with Select via reflection gives us this flexibility (i.e.,
+	// we can either pass the whole array of cases, or the array less the last item).
+	cases := make([]reflect.SelectCase, numChannels + 1)
```
RobAtticus: pretty sure this is not `gofmt`. While I don't like that it is smushed together, this is just going to revert back to that once `gofmt` is run again.

query/benchmarker.go
```diff
@@ -22,45 +22,46 @@ const (
...
 	limit          uint64
 	memProfile     string
+	workers        uint
 	printResponses bool
 	debug          int
 	fileName       string
 
+	// non-flag fields
 	br            *bufio.Reader
+	statProcessor *statProcessor
```
RobAtticus: I prefer the shorter names of `sp` and `c`, pretty common golang-ism to keep them short.

query/scanner.go
```diff
@@ -10,42 +10,50 @@ import (
 // scanner is used to read in Queries from a Reader where they are
 // Go-encoded and then distribute them to workers
 type scanner struct {
-	r     io.Reader
-	limit *uint64
+	reader io.Reader
```
RobAtticus: keep as `r`

query/stat_processor.go
```diff
@@ -9,11 +9,21 @@ import (
```
RobAtticus: not sure any of the changes in this file are needed. it would also help reduce the scope of what I need to review if we remove the subjective refactorings and leave it mostly Clickhouse changes

sunsingerus:
>There are several refactorings that are not related to the Clickhouse code
>...
>some of them seem to change Golang-style short names (1-2 letters even) for longer, descriptive names.
>...
>to revert some of the more subjective changes...

Thanks for the review, and yes, I agree - there are some refactorings not directly related to the ClickHouse functionality. These refactorings originate from the code reversing process - when I've started with implementation, I had very vague understanding of the project, therefore each tool has to be reversed and understood. That's where small code-related refactorings are very handy. As Martin Fowler said: "Whenever I have to think to understand what the code is doing, I ask myself if I can refactor the code to make that understanding more immediately apparent." So through those small refactorings comes code understanding. As to extracting those small refactoring into separated PRs - yes, of course this can be done, but those refactorings are not the goal by themselves and as Martin says: "You refactor because you want to do something else, and refactoring helps you do that other thing."

As a next step, I'll review all the changes, especially subjective ones and revert those that does not makes code better or are not liked by formatter.
I'll notify as soon as update would be ready.

RobAtticus:
Great, thank you! With regards to refactorings, some of them are definitely improvements -- spelling fixes, adding code comments where they are sparse before, and some of the name changes (e.g., `scaleVar ` -> `scale`). However so that the review process does not get too unruly or drawn out, I'd like to focus the PR on Clickhouse stuff.

In general I agree with the gist of what you're saying and definitely want to make the code approachable for others to contribute so we can definitely look at those changes in isolation and see which ones make sense. Just want to keep the PRs more narrowly focused.

cmd/tsbs_generate_queries/databases/clickhouse/devops.go
sunsingerus: Ok, agreed. I've figured out that there is sql-formatter shipped with ClickHouse, which formats queries into the 'canonical form of ClickHouse SQL dialect'  -  and I've used it in new iteration. 
Summary: updated

cmd/tsbs_generate_queries/databases/timescaledb/devops.go
sunsingerus: Ok, agreed. In general, I've rolled back all Timescale changes because it is a rudiment of of code reverse and understanding.
Summary: updated

cmd/tsbs_generate_queries/uses/devops/common.go
sunsingerus: Ok, agreed - gofmt has different view on this part than I am.
Anyway, I've run all code through `gofmt`, so there should not be any issues like this
Summary: updated

cmd/tsbs_generate_queries/uses/devops/common.go
sunsingerus: Ok, agreed.

cmd/tsbs_generate_queries/uses/devops/common.go
sunsingerus: 1. Regarding changing order of the arguments - main idea is that it seems to be more human-friendly to say `get 5 random items out of 12` - and I personally stumbled several times during code digging in this. Totally agree that  this is subjective, but still would be nice (from my point of view) to keep it
2. As to `all the places` - yes, hope to find and modify all the places - there are not so many of them and GUI helps a lot with this.

cmd/tsbs_generate_queries/uses/devops/groupby_orderby_limit.go
sunsingerus: In general, it looks (at least for me) like filling `struct` data type in vertical form correlates with defining `struct` in the same vertical form. At least it seems to be easier to refer to `struct` fields when reviewing code. But, fill-one-field of the struct case can be treated separately - and may be it is easier to fill one field as a one line. However, as to multi-field structs - it really seems better to fill the vertically. So my proposition is to fill one-field structs as one-liners and multi-field - as multi-line.

Summary: this line updated, some other propose to keep multi-lined

cmd/tsbs_load_timescaledb/main.go
sunsingerus: Ok, agreed (and this is one-liner case)

load/creator.go
sunsingerus: More clearly separates methods in the interface, does it not? At least when there are many methods it helps to read easier (for me at least).

load/duplex_channel.go
sunsingerus: Ok, agreed. Code-digging artifact and indeed, `newXXX()` is used all over the project, so there is no need to introduce any `createXXX()`
Summary: updated

cmd/tsbs_load_timescaledb/main.go
sunsingerus: Summary: updated

load/loader.go
sunsingerus: Yes, agreed
Summary: updated

cmd/tsbs_generate_queries/uses/devops/common.go
RobAtticus: Cool, I agree with 1. So if all the spots are gotten (since it is unexported it should only be in that package), then it's a fine change to me.

cmd/tsbs_generate_queries/databases/timescaledb/devops.go
RobAtticus: I don't see these changes yet?

load/scan.go
sunsingerus: Let's just call it `sendBatch`, so there will be two functions in charge of sending batches:
1. `func sendBatch(ch *duplexChannel, count *int, unsent []Batch)`
2. `func sendOrQueueBatch(ch *duplexChannel, count *int, batch Batch, unsent []Batch)`

and I am not sure about correctness of `	*count--` and `*count++` being outside of `if`s in those functions

load/scan.go
sunsingerus: Ok, agreed.
Summary: updated

load/scan.go
sunsingerus: Ok, agreed. `gofmt` did indeed smushed this piece as `numChannels+1`. let it be so
Summary: updated

query/benchmarker.go
sunsingerus: Ok, agreed
Summary: updated

query/scanner.go
sunsingerus: Ok, agreed
Summary: updated

query/stat_processor.go
sunsingerus: Ok, agreed.
Summary: updated

cmd/tsbs_generate_queries/databases/timescaledb/devops.go
sunsingerus: Just pushed, I am writing this notes along with code review

sunsingerus:
PR updated, some files are completely excluded, some other are partially reverted back.

cmd/tsbs_generate_queries/uses/devops/common.go
RobAtticus: Actually if it's not too much trouble, can you separate this out for a separate PR like you did for the other change?

sunsingerus: Yes, think this `getRandomHosts` refactoring can be extracted into separate PR. We may complete it before this big `ClickHouse` PR

sunsingerus: Extracted random refactoring into separate PR [here](https://github.com/timescale/tsbs/pull/27)
Let's move on with this new PR (random functions) and then return back to ClickHouse

sunsingerus:
Looking into code modifications and it seems to me, that there are some more smaller PRs, that can be extracted from this mega-PR (as we did with [random functions](https://github.com/timescale/tsbs/pull/27). This may be the way to go.

RobAtticus:
@sunsingerus You might be right. I think reviewing some things piece meal will go more smoothly.

sunsingerus:
PR updated, one more smaller-PR extracted [as a separate PR](https://github.com/timescale/tsbs/pull/28)

sunsingerus:
Continue to extract separated items into smaller PRs. One more with code cleaning [just added](https://github.com/timescale/tsbs/pull/29)

sunsingerus:
One more smaller PR extract from this mega-PR. Code refactoring and naming clarification [PR](https://github.com/timescale/tsbs/pull/30)

sunsingerus:
Continue to extract small PRs from this mega-PR.
[shell script PR](https://github.com/timescale/tsbs/pull/31)
[code refactoring](https://github.com/timescale/tsbs/pull/32)

cmd/tsbs_generate_queries/databases/clickhouse/devops_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func TestDevopsGetHostWhereWithHostnames(t *testing.T) {
+	cases := []struct {
+		desc      string
+		hostnames []string
+		useTags   bool
+		want      string
+	}{
+		{
+			desc:      "single host - no json or tags",
```
RobAtticus: seems like these test descriptions can be updated to remove mention of `json`

cmd/tsbs_load_clickhouse/creator.go
```diff
@@ -0,0 +1,235 @@
...
+// loader.DBCreator interface implementation
+func (d *dbCreator) Init() {
+	br := loader.GetBufferedReader()
+	d.readDataHeader(br)
+}
+
+// readDataHeader fills dbCreator struct with data structure (tables description)
+// specified at the beginning of the data file
+func (d *dbCreator) readDataHeader(br *bufio.Reader) {
+	// First N lines are header, with the first line containing: "tags:comma-separated list of tags"
```
RobAtticus: This comment leads me to believe the first line is:
```
tags:hostname,region,datacenter,...
```
rather than:
```
tags,hostname,region,datacenter,...
```

(same with the description of second through N-1 lines)

sunsingerus:
>Can you elaborate on whether that is the case, or if you were mostly trying to re-use TimescaleDB scaffolding here? 
>... what the benefits or drawbacks of using those are? ...
>... Since it seems to support a variety of connection methods (HTTP, JDBC, etc), did you see if there were performance differences?

ClickHouse exposes two API endpoints:
1. native binary API endpoint
2. HTTP endpoint

Native binary API is less (should actually be read as poorly) documented and thus very few driver implementations are based on it. Go adapter [used](https://github.com/kshvakov/clickhouse) is one of those native-API implementations. Native protocol is considered to be more effective and faster than HTTP. HTTP - on the other hand - is simpler, well documented and generally more welcomed by application developers. JDBC driver implementation, for example, is based over HTTP protocol under the hood. So, general idea is that native protocol is expected to be faster, however, there are no any numbers available (at my knowledge). This may be one of the possible future test-cases - compare speed of native vs HTTP protocols on the same tasks.

So my thought were like:
1. Seems to be faster
2. Close to existing code - easier to adopt.

cmd/tsbs_generate_queries/databases/clickhouse/devops_test.go
sunsingerus: Yes, agree, there is no JSON support in ClickHouse atm.
Updated.

cmd/tsbs_load_clickhouse/creator.go
```diff
@@ -0,0 +1,238 @@
...
+		i++
+	}
+}
+
+// loader.DBCreator interface implementation
+func (d *dbCreator) DBExists(dbName string) bool {
+	db := sqlx.MustConnect(dbType, getConnectString(false))
+	defer db.Close()
+
+	sql := "SELECT name, engine FROM system.databases WHERE name = '" + dbName + "'"
```
LeeHampton: You mostly use string concatenation for dynamically generating strings in this package, whereas you use string formatting (eg fmt.Sprintf) in the query generation package. Any reason for this discrepancy? I generally find the concatenation harder to read.

cmd/tsbs_load_clickhouse/creator.go
```diff
@@ -0,0 +1,238 @@
...
+	sql := "DROP DATABASE IF EXISTS " + dbName
+	if _, err := db.Exec(sql); err != nil {
+		panic(err)
+	}
+	return nil
+}
+
+// loader.DBCreator interface implementation
+func (d *dbCreator) CreateDB(dbName string) error {
+	// Connecto to ClickHouse in general and CREATE DATABASE
```
LeeHampton: *Connect

cmd/tsbs_load_clickhouse/creator.go
```diff
@@ -0,0 +1,238 @@
...
+		return fmt.Errorf("input header in wrong format. got '%s', expected 'tags'", parts[0])
+	}
+	createTagsTable(db, parts[1:])
+	tableCols["tags"] = parts[1:]
+
+	// d.cols content are lines (metrics descriptions) as:
+	//cpu,usage_user,usage_system,usage_idle,usage_nice,usage_iowait,usage_irq,usage_softirq,usage_steal,usage_guest,usage_guest_nice
+	//disk,total,free,used,used_percent,inodes_total,inodes_free,inodes_used
+	//nginx,accepts,active,handled,reading,requests,waiting,writing
+	for _, cols := range d.cols {
```
LeeHampton: You have a `createTagsTable` function above for the tags table, why not extract a similar one for the metrics tables instead of having all of the logic in this function? Would make this function a bit easier to read (and test).

cmd/tsbs_load_clickhouse/process.go
```diff
@@ -0,0 +1,370 @@
...
+	for _, tagRow := range tagRows {
+		// id of the bew tag
+		id++
+
+		// unfortunately, it is not possible to pass a slice into variadic function of type interface
+		// more details on the item:
+		// https://blog.learngoprogramming.com/golang-variadic-funcs-how-to-patterns-369408f19085
+		// Passing a slice to variadic param with an empty-interface
+		// +1 is here for id
+		var ihellos []interface{} = make([]interface{}, len(tagRow)+1)
```
LeeHampton: Even though ihellos is just a placeholder to wrangle a slice into a variadic function call, I think we should change the variable names from `hello` and `ihellos` here to something that makes a bit more sense in this context ;-)

cmd/tsbs_load_clickhouse/profile.go
```diff
@@ -0,0 +1,49 @@
+package main
```
LeeHampton: So I know we kind of started this pattern by putting the `profile` stuff in `tsbs_load_timescaledb`, but now that we're adding general system profiling to more of the databases, I'm wondering if it might make sense to extract this to a generic package that could be re-used by all of the databases, instead of having to duplicate OS profiling calls for each separate DB that we implement?

cmd/tsbs_run_queries_clickhouse/main.go
```diff
@@ -0,0 +1,160 @@
+// tsbs_run_queries_clickhouse speed tests ClickHouse using requests from stdin or file.
+//
+// It reads encoded Query objects from stdin or file, and makes concurrent requeststo the provided ClickHouse endpoint.
```
LeeHampton: *requests to

cmd/tsbs_load_clickhouse/profile.go
RobAtticus: There's actually a few problems with this profiling, the big one being that it doesn't work if the client & server are not the same. Given that, I think it's probably better to just not include this in new databases (i.e., remove this file and its other code) until we come up with a cleaner way to do this. It probably should be removed from TimescaleDB too but I haven't bothered to do it yet.

cmd/tsbs_load_clickhouse/creator.go
sunsingerus: There is no strict rule, but main idea is that short one-line-strings are constructed using concatenation, while long-multi-line strings are fmt.Sprintf-ed. It is exactly like this in `creator.go` under discussion, where one-line SQL queries (like short `SELECT` or `CREATE DATABASE`) are concatenated and multi-line queries (like `CREATE TABLE long table spec goes here`) are fmt.Sprintf-ed. The threshold after which string must not be concatenated is, I believe, line break. However, this is just personal.

cmd/tsbs_load_clickhouse/creator.go
sunsingerus: Fixed

cmd/tsbs_load_clickhouse/creator.go
sunsingerus: This makes sense

cmd/tsbs_load_clickhouse/creator.go
sunsingerus: Updated to `fmt.Sprintf`

cmd/tsbs_load_clickhouse/creator.go
sunsingerus: Updated

cmd/tsbs_load_clickhouse/process.go
sunsingerus: Yes, this piece of code is not nice. However, it is not easy to name this work-around with something reasonable and descriptive at the same time. Description work-around naming is a challenge by itself. Hope `variadicFriendlyRow` would make a reasonable compromise
Updated

sunsingerus:
```
ihellos := []interface{id}
ihellos = append(ihellos, tagRow...)
```
doesn't work, so no major changes in this part

cmd/tsbs_generate_queries/databases/clickhouse/devops.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+}
+
+// GenerateEmptyQuery returns an empty query.ClickHouse
+func (d *Devops) GenerateEmptyQuery() query.Query {
+	return query.NewClickHouse()
+}
+
+// getHostWhereWithHostnames creates WHERE SQL statement for multiple hostnames.
+// NB 'WHERE' itself is not included, just hostname filter clauses, ready to concatenate to 'WHERE' string
```
RobAtticus: `NB` -> `NOTE:`

cmd/tsbs_load_clickhouse/process.go
```diff
@@ -0,0 +1,369 @@
...
+	id := startId
+	for _, row := range rows {
+		// id of the new tag
+		id++
+
+		// unfortunately, it is not possible to pass a slice into variadic function of type interface
+		// more details on the item:
+		// https://blog.learngoprogramming.com/golang-variadic-funcs-how-to-patterns-369408f19085
+		// Passing a slice to variadic param with an empty-interface
+		var variadicFriendlyRow []interface{} = make([]interface{}, len(row)+1) // +1 here for additional 'id' column value
```
RobAtticus: `variadicArgs` is a shorter name

cmd/tsbs_generate_queries/databases/clickhouse/devops.go
sunsingerus: Ok, updated, just used to write **NB** as shortcut for **nota bene** from students' time. May be it is shorter

cmd/tsbs_load_clickhouse/process.go
sunsingerus: Agree, this is shorter - updated.

sunsingerus:
>Thanks for the explanations of why you chose this.
My pleasure!

>I see this code was used in a blog post (I believe, unless some one else did it as well?),
CTO of Altinity Alexander Zaytsev wrote blogpost with detailed comparison of databases available in the test

>so presumably the performance was satisfactory.
Each of databases has an area where it shines! Timescale and ClickHouse are in general faster than Influx, however Influx simply surprised with disk usage - it seems not to need any disk space at all!

cmd/tsbs_generate_queries/databases/clickhouse/devops.go
```diff
@@ -0,0 +1,341 @@
...
+	meanClauses := make([]string, numMetrics)
+	for i, m := range metrics {
+		meanClauses[i] = "mean_" + m
+		selectClauses[i] = fmt.Sprintf("avg(%s) AS %s", m, meanClauses[i])
+	}
+
+	hostnameField := "hostname"
+	joinClause := ""
+	if d.UseTags {
+		//hostnameField = "tags.hostname"
```
LeeHampton: Let's get rid of this comment

sunsingerus: Yes, this part of code is not clear and looks like some development retries - I do not remember clearly what was that, sorry. Anyway, need to be cleaned.
Updated

sunsingerus:
Thank you for your efforts - really appreciate and enjoying the process!
Two small PRs more are ready - Influx shell script and docs update. And that would be all for now, have no more ready updates at the moment

sunsingerus:
As discussed [here](https://github.com/timescale/tsbs/pull/38#issuecomment-440327521) doc modifications are merged into this PR.
PR updated.

RobAtticus:
Thanks again, merged!

PR #28, opened 2018-11-08
Add limit on generated data points

sunsingerus:
Add useful feature - limit on number of generated points.
So it is possible to say 'generate data points within this timeframe, but not more than X'
Make simulator more human-friendly in terms of naming.


cmd/tsbs_generate_data/main.go
```diff
@@ -173,6 +174,7 @@ func init() {
 	flag.StringVar(&profileFile, "profile-file", "", "File to which to write go profiling data")
 
 	flag.DurationVar(&logInterval, "log-interval", 10*time.Second, "Duration between host data points")
+	flag.Uint64Var(&limit, "limit", 0, "Limit the number of data point to generate, 0 = no limit")
```
LeeHampton: Limit is such an all purpose name for a CLI flag that it might be a bit confusing here, and could lead to weird naming if we ever introduce some other kind of limit. I think it's perfect as an argument/variable name in the specific contexts you've introduced it in, but maybe call it something like `max-data-points` for the CLI flag?

sunsingerus: Yes, think this is a good point - to name CLI parameter more clearly. Ok, let it be `max-data-points`. Let me fix PR accordingly

sunsingerus: updated

sunsingerus:
PR updated according to @LeeHampton 's review

PR #30, opened 2018-11-09
Code refactoring: make query runner cleaner

sunsingerus:
1. Rename --limit CLI flag (which limits number of queries to be run) to --max-queries
The same reason as discussed here: https://github.com/timescale/tsbs/pull/28#discussion_r231954697
2. Rename 'ret' into 'runner', it is much clearer in the code
3. Refactor query.BenchmarkRunner struct to resemble loader.BenchmarkRunner struct for unification
4. Clean and clarify namings


sunsingerus:
>summary should start with an imperative verb

Yes, agree, this makes easier to understand what's going on.
PR updated accordingly.

PR #31, opened 2018-11-12
Add shell script to run test queries for TimescaleDB test case

sunsingerus:
1. Add new script running queries. This runner script uses '--max-queries' CLI flag, which limits number of test queues. Handy when you'd like to make short test session.
2. Update generate_run_script.py to honor '--max-queries' CLI flag, instead of previously used '--limit' flag, which is ambiguous

### DIFF
--- a/scripts/run_queries_timescaledb.sh
+++ b/scripts/run_queries_timescaledb.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# Exit immediately if a command exits with a non-zero status.
+set -e
+
+# Ensure runner is available
+EXE_FILE_NAME=${EXE_FILE_NAME:-$(which tsbs_run_queries_timescaledb)}
+if [[ -z "$EXE_FILE_NAME" ]]; then
+    echo "tsbs_run_queries_timescaledb not available. It is not specified explicitly and not found in \$PATH"
+    exit 1
+fi
+
+# Queries folder
+BULK_DATA_DIR=${BULK_DATA_DIR:-"/tmp/bulk_queries"}
+
+# How many queries would be run
+MAX_QUERIES=${MAX_QUERIES:-"0"}
+
+# How many concurrent worker would run queries - match num of cores, or default to 4
+NUM_WORKERS=${NUM_WORKERS:-$(grep -c ^processor /proc/cpuinfo 2> /dev/null || echo 4)}
+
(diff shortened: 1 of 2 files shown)


scripts/run_queries_timescaledb.sh
```diff
@@ -0,0 +1,46 @@
...
+if [[ -z "$EXE_FILE_NAME" ]]; then
+    echo "tsbs_run_queries_timescaledb not available. It is not specified explicitly and not found in \$PATH"
+    exit 1
+fi
+
+# Queries folder
+BULK_DATA_DIR=${BULK_DATA_DIR:-"/tmp/bulk_queries"}
+
+MAX_QUERIES=${MAX_QUERIES:-"0"}
+NUM_WORKERS=${NUM_WORKERS:-$(grep -c ^processor /proc/cpuinfo)}  # match # of cores - worker per core
```
RobAtticus: This does not work on non-Linux systems (e.g. MacOS or Windows), perhaps have this fall back to a reasonable default (4? 8?) if it fails?

sunsingerus: Yes, I agree that reasonable default would be useful in this case. PR updated accordingly with fallback to 4 workers in case `NUM_WORKERS` is not explicitly specified

RobAtticus: I do not see the change?

sunsingerus: It below, marked line, the whole code snippet looks like this:
```bash
NUM_WORKERS=${NUM_WORKERS:-$(grep -c ^processor /proc/cpuinfo)}  # match # of cores - worker per core
if [ -z "$NUM_WORKERS" ]; then
    # Workers number is not specified. Is /proc/cpuinfo available? Fallback to reasonable default
    NUM_WORKERS=4
fi
```

sunsingerus: Here is link directly [into code](https://github.com/sunsingerus/tsbs/blob/fb051cbfa5dfd01a437e21631cd039a0e382810f/scripts/run_queries_timescaledb.sh#L14)

sunsingerus: Yes, I agree this one-liner is more compact. Nice to learn :+1: 
Code updated

PR #32, opened 2018-11-12
Refactor code to make it more clear and human-friendly.

sunsingerus:
1. Some variables have 'var' suffix which seems to be excessive. Removed in both go and shell code.
2. Some vars that contain 'number of something' amended with 'Num' suffix,
   some vars that contain 'ID of something' amended with 'ID' suffix,
   some vars that contain 'length of something' amended with 'Len' suffix, etc.
3. Multiple comments added and clarified.
4. Some functions renamed to reflect what they actually do.

### DIFF
--- a/load/scan.go
+++ b/load/scan.go
@@ -5,12 +5,12 @@ import (
 	"reflect"
 )
 
-// ackAndMaybeSend adjust the outstanding batches count and potentially sends
-// another batch to the worker via ch. If unsent is non-empty, send. Returns
-// the updated state of unsent
+// ackAndMaybeSend adjust the unsent batches count
+// and sends one batch (if any available) to the worker via ch.
+// Returns the updated state of unsent
 func ackAndMaybeSend(ch *duplexChannel, count *int, unsent []Batch) []Batch {
 	*count--
-	// if there are still batches waiting, send the next
+	// If there are still batches waiting, send the next
 	if len(unsent) > 0 {
 		ch.sendToWorker(unsent[0])
 		if len(unsent) > 1 {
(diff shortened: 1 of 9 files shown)


load/scan.go
```diff
@@ -5,12 +5,12 @@ import (
...
 )
 
-// ackAndMaybeSend adjust the outstanding batches count and potentially sends
-// another batch to the worker via ch. If unsent is non-empty, send. Returns
-// the updated state of unsent
-func ackAndMaybeSend(ch *duplexChannel, count *int, unsent []Batch) []Batch {
+// sendBatch adjust the outstanding batches count
+// and sends one batch (if any available) to the worker via ch.
+// Returns the updated state of unsent
+func sendBatch(ch *duplexChannel, count *int, unsent []Batch) []Batch {
```
RobAtticus: So I still prefer the old name because it is more clear what the function does. It acknowledges that receipt of a 'done' message and possibly (maybe) sends if there is one to send. Without reading the godoc on this, one may get the incorrect impression of what it's doing.

load/scan.go
```diff
@@ -93,82 +95,108 @@ func scanWithIndexer(channels []*duplexChannel, batchSize uint, limit uint64, br
...
+		if ocnt >= olimit {
+			// We have too many outstanding batches, wait until one finishes (i.e. no default)
 			caseLimit--
 		}
 
 		// Only receive an 'ok' when it's from a channel, default does not return 'ok'
 		chosen, _, ok := reflect.Select(cases[:caseLimit])
 		if ok {
-			unsent[chosen] = ackAndMaybeSend(channels[chosen], &ocnt, unsent[chosen])
+			outstandingBatches[chosen] = sendBatch(channels[chosen], &ocnt, outstandingBatches[chosen])
```
RobAtticus: Still prefer `unsent`. Perhaps `unsentBatches`.

load/scan.go
sunsingerus: Ok, I agree, `ackAndMaybeSend` would be more descriptive in this case. It were my attempt to understand code through refactoring. Updated/reverted accordingly

load/scan.go
sunsingerus: Ok, let it be `unsentBatches` then. Still seems useful to add more clarity to var names, so `unsentBatches` looks like clearer to me than simple `unsent`.
Updated accordingly.

sunsingerus:
PR updated

load/scan.go
```diff
@@ -93,82 +95,108 @@ func scanWithIndexer(channels []*duplexChannel, batchSize uint, limit uint64, br
...
 
-	// Current batches (per channel) that are being filled
-	batches := make([]Batch, numChannels)
-	for i := range batches {
-		batches[i] = factory.New()
+	// Batches details
+	// 1. fillingBatches contains batches that are being filled with items from scanner.
+	//    As soon a batch has batchSize items in it, or there is no more items to come, batch moves to
+	// 2. unsentBatches, which contains batches, ready to be sent to worker.
+	//    As soon as there is place in chan to worker, batch is placed into chan
```
LeeHampton: This is a little unclear. Maybe something like:

```
2. unsentBatches contains batches ready to be sent to a worker. 
As soon as a worker's chan is available (i.e., not blocking), the batch is placed onto that worker's chan.
```

sunsingerus: Yes, agree, this part of text is somewhat clumsy. Need to be clarified.
PR updated.

PR #33, opened 2018-11-13
Add ClickHouse data generator.

sunsingerus:
ClickHouse data generator generates data in pseudo-CSV format for ClickHouse test.
It is the same as TimescaleDB data generator - just another entry point into it.

### DIFF
--- a/cmd/tsbs_generate_data/main.go
+++ b/cmd/tsbs_generate_data/main.go
@@ -2,9 +2,10 @@
 //
 // Supported formats:
 // Cassandra CSV format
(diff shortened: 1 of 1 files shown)


RobAtticus:
If that is the end of the non-Clickhouse PRs, I think we can review the rest in the original PR. It makes sense to merge all those changes at once since the data generator is less useful without the other component parts. 

Will close and focus on the PR, unless you strongly disagree.

sunsingerus:
Yes sure we can review the rest as one big PR. I was just thinking about 'small step-by-step' approach

PR #34, opened 2018-11-14
Update README to reflect the name change of the scale-var flag

LeeHampton:
@RobAtticus @sunsingerus Someone in our support channel caught this yesterday -- one place we forgot to update the -scale-var flag name.


sunsingerus:
Yes indeed, documentation is the thing which is often forgotten

LeeHampton:
Turns out there were a few places in the code that also referred to the old var name, namely error strings and comments -- updated the PR.

Issue #35, opened 2018-11-15
Add timescaledb extension to plain postgres db. (maybe an option?)


did-g:
Hi,
 
When benchmarking timescaledb vs pg using the same requests is interesting, but requests with time_bucket need timescaledb extension.

Something like
```

--- a/cmd/tsbs_load_timescaledb/creator.go
+++ b/cmd/tsbs_load_timescaledb/creator.go
@@ -133,8 +133,8 @@ func (d *dbCreator) CreateDB(dbName string) error {
                        dbBench.MustExec(idxDef)
                }
 
+               dbBench.MustExec("CREATE EXTENSION IF NOT EXISTS timescaledb CASCADE")
                if useHypertable {
-                       dbBench.MustExec("CREATE EXTENSION IF NOT EXISTS timescaledb CASCADE")
                        dbBench.MustExec(
                                fmt.Sprintf("SELECT create_hypertable('%s'::regclass, 'time'::name, partitioning_column => '%s'::name, number_partitions => %v::smallint, chu`

```
timescaledb hook add a small overhead but it's trivial.

Regards
Didier

RobAtticus:
This is a good change, and I have wanted to make sure that we can be tested on plain PostgreSQL as we were before (for our blog post). I'll definitely keep this in mind. If you want to make a PR for the small patch you have there and leave the `time_bucket` change for someone else, that's good too.

RobAtticus:
Closing this thanks to PR #49 being merged

PR #40, opened 2018-11-27
Always create timescaledb extension if it's a postgres target

did-g:
Hi,

cf:
https://github.com/timescale/tsbs/issues/35

Regards
Didier


RobAtticus:
Ah sorry, I misread the diff. This change is actually not needed / not the right approach. I will work on the solution that changes time_bucket -> date_trunc so that this change is unnecessary.

did-g:
No problem, writing this one didn't take too much time.

PR #46, opened 2019-01-02
tsbs_load_timescaledb/creator.go: handle case-sensitive columns

### LINKED ISSUE #45: tsbs_load_timescaledb: fails on case-sensitive columns

lilvinz:
closes #45


lilvinz:
Tests still have to be fixed.

RobAtticus:
I'm inclined to say that it is preferable that the schemas for any other use cases adopt the convention of using all lowercase names rather than add the extra quoting that is unnecessary in the current use cases.

RobAtticus:
Closing this pending further objection.

PR #48, opened 2019-01-02
Handle json tags properly on tsbs_load_timescaledb

### LINKED ISSUE #47: tsbs_load_timescaledb: fails to set tags_id when using jsonb tags
https://github.com/timescale/tsbs/blob/master/cmd/tsbs_load_timescaledb/process.go#L78

This seems not to take into account jsonb. The json payload would have to be decoded.

lilvinz:
closes #47

### DIFF
--- a/cmd/tsbs_load_timescaledb/process.go
+++ b/cmd/tsbs_load_timescaledb/process.go
@@ -1,6 +1,7 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


RobAtticus:
Updated PR here: #48

RobAtticus:
Thanks for the PR!

PR #49, opened 2019-01-21
Add support to generate queries for PG

sdressler:
This is an implementation for #35. One can simply use `postgresql` as format. Running the queries can be done with the default tool.

### DIFF
--- a/cmd/tsbs_generate_queries/databases/timescaledb/devops.go
+++ b/cmd/tsbs_generate_queries/databases/timescaledb/devops.go
@@ -9,16 +9,20 @@ import (
 	"github.com/timescale/tsbs/query"
 )
 
+const oneMinute = 60
+const oneHour = oneMinute * 60
+
 // Devops produces TimescaleDB-specific queries for all the devops query types.
 type Devops struct {
 	*devops.Core
-	UseJSON bool
-	UseTags bool
+	UseJSON       bool
+	UseTags       bool
+	UseTimeBucket bool
 }
 
 // NewDevops makes an Devops object ready to generate Queries.
 func NewDevops(start, end time.Time, scale int) *Devops {
-	return &Devops{devops.NewCore(start, end, scale), false, false}
+	return &Devops{devops.NewCore(start, end, scale), false, false, true}
 }
 
 // GenerateEmptyQuery returns an empty query.TimescaleDB
@@ -56,6 +60,14 @@ func (d *Devops) getHostWhereString(nhosts int) string {
 	return d.getHostWhereWithHostnames(hostnames)
 }
 
+func (d *Devops) getTimeBucket(seconds int) string {
+	if d.UseTimeBucket {
+		return fmt.Sprintf("time_bucket('%d seconds', time)", seconds)
+	}
+
+	return fmt.Sprintf("to_timestamp(((extract(epoch from time)::int)/%d)*%d)", seconds, seconds)
+}
+
 func (d *Devops) getSelectClausesAggMetrics(agg string, metrics []string) []string {
 	selectClauses := make([]string, len(metrics))
 	for i, m := range metrics {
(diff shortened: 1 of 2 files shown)


cmd/tsbs_generate_queries/databases/timescaledb/devops.go
```diff
@@ -56,6 +57,14 @@ func (d *Devops) getHostWhereString(nhosts int) string {
 	return d.getHostWhereWithHostnames(hostnames)
 }
 
+func (d *Devops) getTimeBucket(seconds int) string {
+	if d.UseTimeBucket {
+		return fmt.Sprintf("time_bucket('%d seconds', time)", seconds)
+	}
+
+	return fmt.Sprintf("to_timestamp(((extract(epoch from time)::int)/%d)*%d)", seconds, seconds)
```
RobAtticus: I feel like using `date_trunc` is probably the more natural thing to do here, though admittedly that complicates the interface and won't work, e.g., if we decide to queries on 5-min or non-singular intervals. What do you think?

cmd/tsbs_generate_queries/main.go
```diff
@@ -79,10 +79,11 @@ func getGenerator(format string, start, end time.Time, scale int) utils.DevopsGe
 		return mongo.NewDevops(start, end, scale)
 	} else if format == "mongo-naive" {
 		return mongo.NewNaiveDevops(start, end, scale)
-	} else if format == "timescaledb" {
+	} else if format == "timescaledb" || format == "postgresql" {
```
RobAtticus: My only worry here is that now there is no equivalent in `generate_data` which people might find confusing.

cmd/tsbs_generate_queries/main.go
```diff
@@ -79,10 +79,11 @@ func getGenerator(format string, start, end time.Time, scale int) utils.DevopsGe
...
 		tgen := timescaledb.NewDevops(start, end, scale)
 		tgen.UseJSON = timescaleUseJSON
 		tgen.UseTags = timescaleUseTags
+		tgen.UseTimeBucket = format == "timescaledb"
```
RobAtticus: Maybe there should be a `timescaleUseTimeBucket` flag as well so users can benchmark TimescaleDB w/o time_bucket (not sure why you'd want to, but who knows). You could have it default `true` and then this would become `format == "timescaledb" && timescaleUseTimeBucket`

cmd/tsbs_generate_queries/databases/timescaledb/devops.go
sdressler: Think it depends on what to achieve. As long as the benchmark only needs minute/hour precision, I am happy to use `date_trunc`. Otherwise I prefer the timestamp.

RobAtticus: Okay, we'll leave it like this, since it gives us more flexibility in time groupings

cmd/tsbs_generate_queries/databases/timescaledb/devops.go
```diff
@@ -157,6 +168,7 @@ func (d *Devops) GroupByTimeAndPrimaryTag(qi query.Query, numMetrics int) {
         FROM cpu_avg
         %s
         ORDER BY hour, %s`,
+		d.getTimeBucket(60*60),
```
RobAtticus: might be good to extract these (`60*60` and `60`) to `const`s

cmd/tsbs_generate_queries/main.go
RobAtticus: I think we could keep both, but if you don't want to add the format to `generate_data` (which I am still on the fence about), I think it's fine to remove the postgresql format. We can update the README / documentation to make it clear that you need to set `--timescale-use-time-bucket=false` in order to test on PostgreSQL. I think that's reasonable, actually.

sdressler: Makes sense, will update the PR accordingly. Just let me know if you change your opinion wrt `generate_data`.

RobAtticus:
@sdressler just wanted to clarify I am good with removing the `postgresql` format and just adding a `use-time-bucket` flag, in case you were blocking on that.

sdressler:
No worries, I am blocked by time constraints atm.

RobAtticus:
Cool, no problem. Just wanted to make sure I wasn't the blocker.

sdressler:
That is very kind of you :-)

sdressler:
Finally found the time and updated as discussed.

sdressler:
Done. Go ahead, sir.

RobAtticus:
Thanks again, really appreciate it!

PR #50, opened 2019-02-04
Add SiriDB database support

AnjaBruls:
SiriDB can be tested along with all other DBses.


RobAtticus:
Hey, thanks for submitting this. Is this in a fully finished, ready-to-review state? Could you also provide me with a brief overview of the approach as I am not too familiar with SiriDB so it would help me know what to look for.

AnjaBruls:
Hey @RobAtticus ! Yes it is ready to be reviewed. SiriDB is a fully open-sourced time series database. Its approach is most similar to InfluxDB. This blog (https://siridb.net/blog/how-we-store-time-series-in-siridb/ ) gives a brief overview of how SiriDB is set up. This link (https://siridb.net/documentation/) will provide you with documentation about the query language. The source code of SiriDB can be found on https://github.com/siridb under *SiriDB-server*. On GitHub you can also find a HTTP API, and a terminal client to communicate with SiriDB and a couple of native SiriDB connectors . Hope this answers your question. Thanks in advance for reviewing SiriDB!

cmd/tsbs_generate_data/serialize/siridb.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"io"
+	"log"
+	"strconv"
+
+	qpack "github.com/transceptor-technology/go-qpack"
+)
+
+// SiriDBSerializer writes a Point in a serialized form for TimescaleDB
+type SiriDBSerializer struct{}
```
RobAtticus: This file has several left over/outdated comments referring to TimescaleDB rather that the actual form of the data stored.

cmd/tsbs_generate_data/serialize/siridb.go
```diff
@@ -0,0 +1,70 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		postQpack := len(line)
+
+		binary.LittleEndian.PutUint16(line[indexLenData:], uint16(postQpack-preQpack))
+		metricCount++
+	}
+
+	binary.LittleEndian.PutUint16(line[0:], uint16(metricCount))
```
RobAtticus: Probably unnecessary but I would make each of these at least 32 to be on the safe side.

cmd/tsbs_generate_data/serialize/siridb.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Serialize writes Point p to the given Writer w, so it can be
+// loaded by the TimescaleDB loader. The format is CSV with two lines per Point,
+// with the first row being the tags and the second row being the field values.
+//
+// e.g.,
+// tags,<tag1>,<tag2>,<tag3>,...
+// <measurement>,<timestamp>,<field1>,<field2>,<field3>,...
+func (s *SiriDBSerializer) Serialize(p *Point, w io.Writer) error {
+	line := make([]byte, 4, 1024)
```
RobAtticus: an explainer comment that the first four bytes will be filled in at the end might be helpful

AnjaBruls:
The requested changes are made: the comments are updated for `serialize/siridb.go` as well as for `devops.go`. Also we replaced 16 with 32 bit. 
And sorry for the import-order-changes. I think the go-linter had done this, not sure why. But these files are removed.

cmd/tsbs_generate_queries/databases/siridb/devops.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	combinedMetricsClause := strings.Join(metricsClauses, "|")
+	return "(" + combinedMetricsClause + ")"
+}
+
+const goTimeFmt = "2006-01-02 15:04:05Z"
+
+// GroupByTime selects the MAX for numMetrics metrics in the group `cpu`,
+// per minute for nhosts hosts,
+// e.g. in psuedo-SQL:
```
RobAtticus: I just fixed this in other files, but `pseudo` is spelled wrong here. Sorry :(

cmd/tsbs_load_siridb/main.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"bufio"
+	"flag"
+	"log"
+
+	"github.com/timescale/tsbs/load"
+)
+
+const (
+	account       = "sa"
```
RobAtticus: since these constants are used in `creator.go`, they should probably live there

cmd/tsbs_generate_queries/databases/siridb/devops.go
```diff
@@ -0,0 +1,175 @@
...
+	interval := d.Interval.RandWindow(timeRange)
+	metrics := devops.GetCPUMetricsSlice(numMetrics)
+	whereMetrics := d.getMetricWhereString(metrics)
+	whereHosts := d.getHostWhereString(nHosts)
+
+	humanLabel := fmt.Sprintf("SiriDB %d cpu metric(s), random %4d hosts, random %s by 1m", numMetrics, nHosts, timeRange)
+	humanDesc := fmt.Sprintf("%s: %s", humanLabel, interval.StartString())
+
+	var siriql string
+	if numMetrics == 1 {
```
RobAtticus: I find this part a bit weird, but I haven't dived into the documentation for the sql query language. I'm a bit curious why the case of 1 metric would be more verbose than multiple metrics.

RobAtticus:
No worries, it seems the linters we use are doing different ordering and it's unclear why. Oh well :)

PR looks mostly good to me (thanks!), just a few minor things.

cmd/tsbs_generate_queries/databases/siridb/devops.go
AnjaBruls: The reason for this lies in how the time series are stored. We have stored the series per metric and per host. It is possible to return the series of e.g. 10 metrics  and 5 hosts (using regular expressions), this will return 50 time series in total. If we then merge these series, these 50 series are merged to 1 series. So merged over both host and metric. Unfortunately it is not possible yet to merge per metric for multiple metrics within one query request. Therefore if we want to merge only over host, we have to select one metric at a time. In cases like ` single-groupby-5-8-1 ` we chose to return all the series separate instead and no merging is done.

cmd/tsbs_generate_queries/databases/siridb/devops.go
AnjaBruls: Oh fixed it:)

cmd/tsbs_load_siridb/main.go
AnjaBruls: Yes I agree, I moved them

AnjaBruls:
Great, thanks! I made the requested changes and I synced my fork with the upstream. But I see that the travis-ci build was cancelled. I’m not sure why. Could you try doing a retry of this test? 

Furthermore SiriDB seems to perform well.:) Here are some graphs that show some benchmark results on my system. Data was generated for 30 hosts and the insert and query tests were performed using 1 worker.

querytime

insertspeed

RobAtticus:
Thanks @AnjaBruls, one more thing before I merge: have you tried with datasets with more than 30 hosts (like 1000?). Just want to make sure nothing breaks when you go about 10s of devices to hundreds or thousands.

Thanks, will merge when Travis finishes (I don't know why it stopped).

AnjaBruls:
Thanks for the review!:)

We didn't perform tests for 100s or 1000s of hosts yet. We will do a couple of tests and I let you now when results are in.

AnjaBruls:
Hi @RobAtticus , we performed some test with a 1000 hosts and they went well. Only thing we run into was  that for some queries we had to increase the maximum points which can be returned by a select query. Therefore we added this feature.

RobAtticus:
Thanks again for this PR! Appreciate your patience, it has now been merged.

AnjaBruls:
Great! Thanks for adding SiriDB to the list!

PR #56, opened 2019-02-25
Separate db create from schema create

niksajakovljevic:
There are some cases when database needs to be configured/setup in a certain way (eg. setup db using some other tooling). So preferably you'd only want to drop the metrics table and create a new one and leave the rest of db intact.

### DIFF
--- a/cmd/tsbs_load_timescaledb/creator.go
+++ b/cmd/tsbs_load_timescaledb/creator.go
@@ -76,75 +76,80 @@ func (d *dbCreator) CreateDB(dbName string) error {
 	db := sqlx.MustConnect(dbType, d.connStr)
 	db.MustExec("CREATE DATABASE " + dbName)
 	db.Close()
+	return nil
+}
 
-	dbBench := sqlx.MustConnect(dbType, getConnectString())
-	defer dbBench.Close()
+func (d *dbCreator) PostCreateDB(dbName string) error {
+	if createMetricsTable {
+		dbBench := sqlx.MustConnect(dbType, getConnectString())
+		defer dbBench.Close()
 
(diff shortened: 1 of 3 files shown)


cmd/tsbs_load_timescaledb/creator.go
```diff
@@ -76,75 +76,80 @@ func (d *dbCreator) CreateDB(dbName string) error {
 	db := sqlx.MustConnect(dbType, d.connStr)
 	db.MustExec("CREATE DATABASE " + dbName)
 	db.Close()
+	return nil
+}
 
-	dbBench := sqlx.MustConnect(dbType, getConnectString())
-	defer dbBench.Close()
+func (d *dbCreator) PostCreateDB(dbName string) error {
+	if createFactsTable {
```
RobAtticus: `createMetricsTable` maybe?

cmd/tsbs_load_timescaledb/creator.go
```diff
@@ -76,75 +76,80 @@ func (d *dbCreator) CreateDB(dbName string) error {
...
+		createTagsTable(dbBench, parts[1:])
+		tableCols[tagsKey] = parts[1:]
+
+		for _, cols := range d.cols {
+			parts = strings.Split(strings.TrimSpace(cols), ",")
+			hypertable := parts[0]
+			partitioningField := tableCols[tagsKey][0]
+			tableCols[hypertable] = parts[1:]
+
+			psuedoCols := []string{}
```
RobAtticus: can you fix my spelling mistake? :) `pseudoCols`

RobAtticus:
Also, please add an extended commit description of what the purpose of this change is/why.

niksajakovljevic:
@RobAtticus Sure, I'll add commit description

Issue #39, opened 2018-11-26
Strange decision to remove dbname from connection string in dbCreator initialization routine


maksm90:
I have noticed that `tsbs_load_timescaledb` cannot connect to database using specific username and dbname distinct from this name, e.g., using the following connection string `dbname=postgres user=user`. The problem lies in the following fragment of code _cmd/tsbs_load_timescaledb/creator.go:22_ :
```golang
// Needed to connect to user's database in order to drop/create db-name database
re := regexp.MustCompile(`(dbname)=\S*\b`)
d.connStr = strings.TrimSpace(re.ReplaceAllString(d.connStr, ""))
```

AFAIU the decision to remove dbname from connection string was based on assumption to distinguish the dbname for connection from the working dbname for benchmark. Instead of deletion I propose to check that these names obtained from connection string and option `-db-name` correspondingly are different.

RobAtticus:
That makes sense, we can definitely add a flag like `-conn-db` to specify the database name to connect to for original creation. Did not think of that use case, but good point.

antekresic:
I believe we can close this one since it was resolved by PR #60.

maksm90:
Ok, thanks

PR #60, opened 2019-03-05
Added connection DB flag to tsbs_load_timescaledb

antekresic:


RobAtticus:
This functionality is already exposed via the `--db-name` flag, no?:
```
$ tsbs_load_timescaledb --help
Usage of tsbs_load_timescaledb:
...
  -db-name string
    	Name of database (default "benchmark")
...
```

(The flag is defined/exists in the `load` pkg since it is shared across loaders)

RobAtticus:
Ah I think this is supposed to solve a different problem -- connecting when you want to create the original database? May need to work on some naming issues here.

antekresic:
Yeah, this is regarding issue #39. I thought that as well at first but I guess people need something like this when connecting to the database where they have to use a custom database name. 

How about `--custom-conn-db`? Or should we change the other flag name to make it more distinct from this one?

RobAtticus:
Let's go with `--admin-db-name` and a description like:
```
Database to connect to in order to create additional benchmark databases. By default this is the same as the `db-user` (i.e., `postgres` if neither is set), but sometimes a user does not have its own database.
```
A bit of a mouthful, but gets it across. Then maybe update the commit message to give an extended explaining the reason/need for this change.

Appreciate the PR!

antekresic:
Updated PR based on the recent comment.

Let me know if there is anything else that needs changing.

Thanks!

RobAtticus:
Hi @antekresic, I think you need to rebase your branch on the latest (rather than merge ours into yours) because now there are some duplicated commits in your PR. And then squash all your commits into one.

antekresic:
Hey @RobAtticus 

Sorry about that, I took a wrong turn rebasing the branch previously. I believe its fixed now.

PR #63, opened 2019-03-28
Add binary format support

niksajakovljevic:
Using binary format when talking to TimescaleDB
means less data being sent back and forth.
Config option is added to force TEXT format if needed
(binary is default). PGX driver is used for
binary and PQ driver for TEXT.

### DIFF
--- a/cmd/tsbs_load_timescaledb/main.go
+++ b/cmd/tsbs_load_timescaledb/main.go
@@ -13,14 +13,14 @@ import (
 	"sync"
 	"time"
 
-	_ "github.com/lib/pq"
 	"github.com/timescale/tsbs/load"
 )
 
 const (
-	dbType       = "postgres"
 	timeValueIdx = "TIME-VALUE"
 	valueTimeIdx = "VALUE-TIME"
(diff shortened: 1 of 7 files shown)


cmd/tsbs_load_timescaledb/main.go
```diff
@@ -167,5 +176,10 @@ func getConnectString() string {
 		connectString = fmt.Sprintf("%s password=%s", connectString, pass)
 	}
 
+	if forceTextFormat {
+		// we assume we're using pq driver
```
perrygeo: You might be able to force text format using the pgx driver by setting `PreferSimpleProtocol`to true in the pgx connection config? Probably not worth changing at this point but maybe if we have more time.

niksajakovljevic: Yeah, I though about that. However, copy protocol in PGX only supports binary so we would still need PG...

niksajakovljevic:
@RobAtticus I'll update the commit message with my findings. In my benchmarks I've got around 5-10% increase in write throughput and about 35% improvement for queries when using binary format.

PR #65, opened 2019-04-09
Rewrite tsbs_generate_queries to use internal/inputs

RobAtticus:
This PR continues on the work in the previous one that changed
tsbs_generate_data to use a new internal/inputs package. This PR
adds a new Generator type for query generation called QueryGenerator.

Now that these two generators share some common code, they both
become much more robust and easier to test and manage. Previously
tsbs_generate_queries had no test coverage, but with this change it
will actually have quite high coverage.

There are still some rough spots with this refactor. In particular,
how the useCaseMatrix is handled needs some more thought, especially
if we are going to add more use cases going forward. Additionally,
the database specific flags like TimescaleUseJSON could probably
be handled in a cleaner way as well.

### DIFF
--- a/internal/inputs/generator_queries.go
+++ b/internal/inputs/generator_queries.go
@@ -1,0 +1,272 @@
+package inputs
+
+import (
+	"bufio"
+	"encoding/gob"
+	"flag"
+	"fmt"
+	"io"
+	"math/rand"
+	"os"
+	"sort"
+	"time"
+
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/databases/cassandra"
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/databases/clickhouse"
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/databases/influx"
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/databases/mongo"
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/databases/siridb"
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/databases/timescaledb"
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/utils"
+)
+
+// Error messages when using a QueryGenerator
+const (
+	ErrInvalidQueryConfig = "invalid config: QueryGenerator needs a QueryGeneratorConfig"
+	ErrEmptyQueryType     = "query type cannot be empty"
+
+	errBadQueryTypeFmt        = "invalid query type for use case '%s': '%s'"
+	errCouldNotDebugFmt       = "could not write debug output: %v"
+	errCouldNotEncodeQueryFmt = "could not encode query: %v"
+	errCouldNotQueryStatsFmt  = "could not output query stats: %v"
+)
+
+// QueryGeneratorConfig is the GeneratorConfig that should be used with a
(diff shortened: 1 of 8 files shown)


internal/inputs/generator_queries.go
```diff
@@ -0,0 +1,272 @@
...
+// different querying techniques, e.g. SQL or REST), but is consumed by TSBS
+// query runners like tsbs_run_queries_timescaledb.
+type QueryGenerator struct {
+	// Out is the writer where data should be written. If nil, it will be
+	// os.Stdout unless File is specified in the GeneratorConfig passed to
+	// Generate.
+	Out io.Writer
+	// Out2 is where non-generated data should be written. If nil, it will be
+	// os.Stderr.
+	Out2 io.Writer
```
LeeHampton: I feel like this might be a little annoying to read for newcomers. Maybe something like `DataOutput` and `LogOutput` or `InfoOutput` ?

internal/inputs/generator_queries.go
```diff
@@ -0,0 +1,272 @@
...
+
+	if g.Out2 == nil {
+		g.Out2 = os.Stderr
+	}
+
+	return nil
+}
+
+func (g *QueryGenerator) getUseCaseGenerator(c *QueryGeneratorConfig) (utils.DevopsGenerator, error) {
+	var ret utils.DevopsGenerator
```
LeeHampton: This is probably a holdover from the older incarnation of this tool, but I always find vars with names like `ret` super hard to follow, might be good to change them to `gen` or `generator` while we're here.

internal/inputs/generator_queries.go
```diff
@@ -0,0 +1,272 @@
...
+// QueryGenerator. It includes all the fields from a BaseConfig, as well as
+// options that are specific to generating the queries to test against a
+// database, such as the query type and individual database options.
+type QueryGeneratorConfig struct {
+	BaseConfig
+	QueryType            string
+	InterleavedGroupID   uint
+	InterleavedNumGroups uint
+
+	// TODO - I think this needs some rethinking, but a simple, elegant solution escapes me right now
```
LeeHampton: Hm, yeah, I'm drawing blanks on a clever solution here as well. But I do see this becoming a pain point as we add more DBs and additional functionality for queries that get generated. One possibility is to have a single `QueryConfig` field that is defined as an `interface{}` with functions that type infer it to a specific database's configuration when needed for validation or generation, but I suppose we'll cross that bridge when we get there.

internal/inputs/generator_queries_test.go
```diff
@@ -0,0 +1,532 @@
...
+	}
+	if got := useGen.(*timescaledb.Devops).UseJSON; got != c.TimescaleUseJSON {
+		t.Errorf("timescaledb UseJSON not set correctly: got %v want %v", got, c.TimescaleUseJSON)
+	}
+	if got := useGen.(*timescaledb.Devops).UseTimeBucket; got != c.TimescaleUseTimeBucket {
+		t.Errorf("timescaledb UseTimeBucket not set correctly: got %v want %v", got, c.TimescaleUseTimeBucket)
+	}
+
+	c.TimescaleUseJSON = true
+	c.TimescaleUseJSON = true
```
LeeHampton: Looks like this is a duplicate line, or maybe we're just being extra safe ;-)

PR #70, opened 2019-04-24
Fix parsing columns when --do-create-db=false

LeeHampton:


LeeHampton:
@RobAtticus Before I write a series of new tests for this, I wanted to make sure this refactor looks like a decent path to go down to you.

Basically, I've split the PostCreateDB method into 2 different methods, `getFieldAndIndexDefinitions` definitions for parsing the generated data headers and coming up with field and index definitions, and `createTableAndIndexes` for running all of the DB commands based on the field/index definitions and user settings. 

This makes it easier to make proper use of the `createMetricsTable` flag (the bug this ticket actually fixes), and it should make things more testable. I've also renamed a bunch of the old fields to be a little more clear (no more `parts` variable name...).

RobAtticus:
Yup this looks reasonable to me. Separate from this PR we should try and clean up the global state -- seems to be a bit of a mess to keep track of.

LeeHampton:
Definitely agreed on that one @RobAtticus

LeeHampton:
@RobAtticus Added tests. This should be ready for review now.

LeeHampton:
Good call @RobAtticus -- done.

PR #72, opened 2019-05-16
Add unit tests to increase code coverage for query generation

antekresic:
Making a small PR for unit tests to give people a chance to review and comment on testing style. 

Once this one gets approved and merged, expect a bigger PR covering more code using the same style.

### DIFF
--- a/cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
+++ b/cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
@@ -6,6 +6,9 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/uses/devops"
+	"github.com/timescale/tsbs/query"
 )
 
 func TestDevopsGetHostWhereWithHostnames(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +154,333 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+		expectedHumanDesc  string
+		expectedHypertable string
+		expectedSQLQuery   string
+	}{
+		{
+			desc:               "no JSON or tags",
+			expectedHumanLabel: "TimescaleDB last row per host",
+			expectedHumanDesc:  "TimescaleDB last row per host",
+			expectedHypertable: "cpu",
+			expectedSQLQuery:   "SELECT DISTINCT ON (hostname) * FROM cpu ORDER BY hostname, time DESC",
```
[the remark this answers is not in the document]
blagojts: And with queries that are not constant but have specific parts that are filled in, I suggest that we keep the fixed parts as templates in constant, and then format them with the generated values. 

This way the tests will be a lot shorter and all the queries will be in one place in one file for a quick review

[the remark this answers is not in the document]
antekresic: My reasoning was keeping the same style of test for all the query generation functions. Thats why I'm testing a fixed string (which could change to a dynamic one in the future).

As for making the tests shorter by reusing the templates, I guess you would have to basically rewrite all the logic in the function by reusing the same functions used to fill out the template and the test would be mirroring the function in test. I'm not fond of that approach since you are just writing the same logic twice and checking that it generates the same output. I decided to go this way so we can at least see the whole query visually in the test to check for sanity.

I, also, wanted to make a point of not changing the current code until we have some test coverage. After that we can introduce some refactoring at which point we can refactor the tests if there is a need for it.

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +154,333 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+			verifyQuery(t, q, c.expectedHumanLabel, c.expectedHumanDesc, c.expectedHypertable, c.expectedSQLQuery)
+		})
+	}
+}
+
+func verifyQuery(t *testing.T, q query.Query, humanLabel, humanDesc, hypertable, sqlQuery string) {
+	tsq, ok := q.(*query.TimescaleDB)
+
+	if !ok {
+		t.Error("Filled query is not *query.TimescaleDB type")
```
RobAtticus: can probably just do `t.Fatal` here and skip the return

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +154,333 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+		return
+	}
+
+	if string(tsq.HumanLabel) != humanLabel {
```
RobAtticus: personally a fan of the form:
```
if got := string(tsq.HumanLabel); got != humanLabel {
  t.Errorf("incorrect HumanLabel: got\n%s\nwant\n%s", got, humanLabel)
}
```

Maybe you can break the printing into a few lines, but I usually just do one. I would prefer to use the `got`/`want` naming instead of `got`/`expected` since we use `want` everywhere else

cmd/tsbs_generate_queries/uses/devops/common_test.go
```diff
@@ -37,6 +37,26 @@ func TestNewCoreEndBeforeStart(t *testing.T) {
...
+	c := NewCore(s, e, scale)
+
+	rand.Seed(100) // Resetting seed to get a deterministic output.
+	coreHosts := strings.Join(c.GetRandomHosts(n), ",")
+
+	rand.Seed(100) // Resetting seed to get a deterministic output.
+	randomHosts := strings.Join(getRandomHosts(n, scale), ",")
+
+	if coreHosts != randomHosts {
+		t.Errorf("GetRandomHosts did not return correct output, expected: %s, got %s", randomHosts, coreHosts)
```
RobAtticus: same comment as below

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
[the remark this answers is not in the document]
RobAtticus: Yeah I do like the approach for now of having the whole query, even if it is a bit cluttered. I think @blagojts 's suggestions for that function are good ones for a follow up PR (since as @antekresic mentioned, we will then have tests to cover it in case the refactor goes wrong). Those changes being making the strings `const`s and not use `Sprintf` for no reason.

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +154,333 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+        SELECT hour, tags->>'hostname', mean_usage_user
+        FROM cpu_avg
+        JOIN tags ON cpu_avg.tags_id = tags.id
+        ORDER BY hour, tags->>'hostname'`,
+		},
+	}
+
+	rand.Seed(123) // Setting seed for testing purposes.
+	s := time.Unix(0, 0)
+	e := s.Add(13 * time.Hour)
```
RobAtticus: Just to make this slightly more resilient, I think you should do it as something like:
```
e := s.Add(devops.DoubleGroupByDuration).Add(time.Hour)
```

Otherwise we might have subtle failures if that constant changes that are hard to track down. Same for other test funcs

cmd/tsbs_generate_queries/uses/devops/common_test.go
```diff
@@ -100,6 +120,24 @@ func TestGetCPUMetricsSlice(t *testing.T) {
 	}
 }
 
+func TestGetAllCPUMetrics(t *testing.T) {
+	result := strings.Join(GetAllCPUMetrics(), ",")
+	metrics := strings.Join(cpuMetrics, ",")
+
+	if result != metrics {
+		t.Errorf("GetAllCPUMetrics returned wrong output, expected: %s, got: %s", metrics, result)
```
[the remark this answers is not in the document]
antekresic: Updated to match consistency. I personally use `assert` package for these so I don't have to worry as much about it.

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
antekresic: Updated to recommendation.

cmd/tsbs_generate_queries/uses/devops/common_test.go
antekresic: Updated.

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
antekresic: Fixed.

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
antekresic: Fixed.

antekresic:
Thanks for the feedback @blagojts and @RobAtticus . 

I incorporated your suggestions and force pushed the commit with matching extended description.

Let me know if there is anything else blocking us from merging this.

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +155,324 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+			expectedHumanDesc:  "TimescaleDB last row per host",
+			expectedHypertable: "cpu",
+			expectedSQLQuery: "SELECT DISTINCT ON (t.hostname) * FROM tags t INNER JOIN LATERAL(SELECT * FROM cpu c " +
+				"WHERE c.tags_id = t.id ORDER BY time DESC LIMIT 1) AS b ON true ORDER BY t.hostname, b.time DESC",
+		},
+	}
+
+	rand.Seed(123) // Setting seed for testing purposes.
+	s := time.Unix(0, 0)
+	e := s.Add(devops.DoubleGroupByDuration).Add(time.Hour)
```
RobAtticus: this doesn't need to be `Add(devops.DoubleGroupByDuration)`, I actually just meant that for the double groupby queries :)

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +155,324 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+        max(usage_user) as max_usage_user, max(usage_system) as max_usage_system, max(usage_idle) as max_usage_idle, ` +
+		"max(usage_nice) as max_usage_nice, max(usage_iowait) as max_usage_iowait, max(usage_irq) as max_usage_irq, " +
+		"max(usage_softirq) as max_usage_softirq, max(usage_steal) as max_usage_steal, max(usage_guest) as max_usage_guest, " +
+		`max(usage_guest_nice) as max_usage_guest_nice
+        FROM cpu
+        WHERE (hostname = 'host_9') AND time >= '1970-01-01 02:16:22.646325 +0000' AND time < '1970-01-01 10:16:22.646325 +0000'
+        GROUP BY hour ORDER BY hour`
+	rand.Seed(123) // Setting seed for testing purposes.
+	s := time.Unix(0, 0)
+	e := s.Add(devops.DoubleGroupByDuration).Add(time.Hour)
```
RobAtticus: this should be `devops.MaxAllDuration`

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +155,324 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+	expectedSQLQuery := `SELECT time_bucket('60 seconds', time) AS minute, max(usage_user)
+        FROM cpu
+        WHERE time < '1970-01-01 07:16:22.646325 +0000'
+        GROUP BY minute
+        ORDER BY minute DESC
+        LIMIT 5`
+
+	rand.Seed(123) // Setting seed for testing purposes.
+	s := time.Unix(0, 0)
+	e := s.Add(devops.DoubleGroupByDuration).Add(time.Hour)
```
RobAtticus: This doesn't need to `Add(devops.DoubleGroupByDuration)`, it could just N hours

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +155,324 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+	expectedHypertable := "cpu"
+	expectedSQLQuery := `SELECT time_bucket('60 seconds', time) AS minute,
+        max(usage_user) as max_usage_user
+        FROM cpu
+        WHERE (hostname = 'host_9') AND time >= '1970-01-01 10:05:50.646325 +0000' AND time < '1970-01-01 10:05:51.646325 +0000'
+        GROUP BY minute ORDER BY minute ASC`
+
+	rand.Seed(123) // Setting seed for testing purposes.
+	s := time.Unix(0, 0)
+	e := s.Add(devops.DoubleGroupByDuration).Add(time.Hour)
```
RobAtticus: This doesn't need to Add(devops.DoubleGroupByDuration), it could just N hours

cmd/tsbs_generate_queries/databases/timescaledb/devops_test.go
```diff
@@ -152,3 +155,324 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+			expectedHypertable: "cpu",
+			expectedSQLQuery: "SELECT * FROM cpu WHERE usage_user > 90.0 and time >= '1970-01-01 00:08:59.080812 +0000'" +
+				" AND time < '1970-01-01 12:08:59.080812 +0000' AND (hostname = 'host_5' OR hostname = 'host_9' " +
+				"OR hostname = 'host_1' OR hostname = 'host_7' OR hostname = 'host_2')",
+		},
+	}
+
+	rand.Seed(123) // Setting seed for testing purposes.
+	s := time.Unix(0, 0)
+	e := s.Add(devops.DoubleGroupByDuration).Add(time.Hour)
```
RobAtticus: this should be `devops.HighCPUDuration`

antekresic:
@RobAtticus fixed all the durations. Should be good to go now.

PR #73, opened 2019-05-17
Add Akumuli database support

Lazin:


RobAtticus:
Hi @Lazin , I see you've updated this (though it does need a few more conflicts fixed). I will try to look into this this week so we can help move it forward, now that CrateDB is merged in.

Lazin:
Looking forward to it! I fixed merge conflicts but it looks like something is broken on CI now. I'll investigate this tomorrow.

Lazin:
Hi @RobAtticus , 
I updated and tested PR. It can be reviewed now.

blagojts:
Hello @Lazin 
The akumuli_test.go file contains some errors (tag values can be any type now, no need to have strings only, and the Point struct accepts an interface{} now)

Lazin:
Hi @blagojts 
I updated akumuli_test.go and merged with master.

cmd/tsbs_load_akumuli/creator.go
```diff
@@ -0,0 +1,32 @@
+package main
```
blagojts: If I understood correctly, Akumuli does not have the concept of databases and schemas like PostgreSQL or InfluxDB has. And the db creator is not needed to create a database. But the Creator also handles cleaning of old/stale data (check out tsbs_load_timescaledb/creator.go). I would also change it so that the `DBExists` method always returns `true` so the `RemoveOldDb` gets called by tsbs (where you can put the cleanup code). The calling of RemoveOldDb is controlled by the `do-create-db` flag, which by default is true, but if you want to keep the old data you can set it to false

cmd/tsbs_generate_data/serialize/akumuli.go
```diff
@@ -0,0 +1,133 @@
...
+	bookClosed bool
+	deferred   []byte
+	index      uint32
+}
+
+// Serialize writes Point data to the given writer, conforming to the
+// AKUMULI RESP protocol.  Serialized adds extra data to guide data loader.
+//
+func (s *AkumuliSerializer) Serialize(p *Point, w io.Writer) (err error) {
+	if s.book == nil {
```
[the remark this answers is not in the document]
Lazin: Done

cmd/tsbs_generate_data/serialize/akumuli_test.go
```diff
@@ -0,0 +1,115 @@
...
+	buf := new(bytes.Buffer)
+	for _, point := range points {
+		serializer.Serialize(point, buf)
+	}
+
+	got := buf.String()
+
+	for _, c := range cases {
+		actualCnt := strings.Count(got, c.expValue)
+		if actualCnt != c.expCount {
```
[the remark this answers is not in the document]
Lazin: It's checked on line 73. If one of the test samples is not present in the output required number of times it will fail.

cmd/tsbs_generate_queries/databases/akumuli/devops.go
```diff
@@ -0,0 +1,294 @@
+package akumuli
```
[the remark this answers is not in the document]
Lazin: It's not possible to implement using Akumuli query API. Date/time truncation is not supported.

cmd/tsbs_load_akumuli/creator.go
Lazin: Yes, it doesn't have a notion of database/table/schema. And it's not possible to clean up using the database API. You have to run command in terminal to crate or delete the database.

Lazin:
@blagojts thanks, I appreciated the feedback!
To run the test you should use recent docker image. 0.7.45 is way too old. The latest tag 0.8.69-skylake should work fine.

RobAtticus:
Hey @Lazin , we're seeing failures in CI:
```
# github.com/timescale/tsbs/cmd/tsbs_load_akumuli
cmd/tsbs_load_akumuli/main.go:33:34: not enough arguments in call to load.GetBenchmarkRunner
	have ()
	want (load.BenchmarkRunnerConfig)
```
and
```
# github.com/timescale/tsbs/cmd/tsbs_run_queries_akumuli
cmd/tsbs_run_queries_akumuli/main.go:26:35: not enough arguments in call to query.NewBenchmarkRunner
	have ()
	want (query.BenchmarkRunnerConfig)
```

Lazin:
Hi @RobAtticus,
I fixed this errors and ran all benchmark apps to check if anything else was broken. It all worked for me. Also, updated the doc since the command line interface changed slightly.

cmd/tsbs_generate_data/serialize/akumuli.go
```diff
@@ -0,0 +1,136 @@
...
+// Serialize writes Point data to the given writer, conforming to the
+// AKUMULI RESP protocol.  Serializer adds extra data to guide data loader.
+// This function writes output that contains binary and text data in RESP format.
+func (s *AkumuliSerializer) Serialize(p *Point, w io.Writer) (err error) {
+	deferPoint := false
+
+	buf := make([]byte, 0, 1024)
+	// Add cue
+	const HeaderLength = 8
+	buf = append(buf, "AAAAFFEE+"...)
```
RobAtticus: can you make this a named constant? I am unclear what this magic value is.

cmd/tsbs_generate_data/serialize/akumuli.go
```diff
@@ -0,0 +1,136 @@
...
+				return err
+			}
+			buf = buf[:HeaderLength]
+			buf = append(buf, fmt.Sprintf(":%d", id)...)
+			binary.LittleEndian.PutUint32(buf[:4], id)
+		} else {
+			// Shortcut
+			s.index++
+			tmp := make([]byte, 0, 1024)
+			tmp = append(tmp, "AAAAFFEE*2\n"...)
```
RobAtticus: same thing as above

cmd/tsbs_generate_data/serialize/akumuli.go
```diff
@@ -0,0 +1,136 @@
...
+			buf = append(buf, '+')
+		}
+		buf = fastFormatAppend(v, buf)
+		buf = append(buf, '\n')
+	}
+
+	// Update cue
+	binary.LittleEndian.PutUint16(buf[4:6], uint16(len(buf)))
+	binary.LittleEndian.PutUint16(buf[6:HeaderLength], uint16(len(p.fieldValues)))
+	if deferPoint {
```
RobAtticus: this is probably more go-like written as:
```go
if deferPoint {
  s.deferred = append(s.deferred, buf...)
  return nil
}
_, err = w.Write(buf)
return err
```

cmd/tsbs_generate_data/serialize/akumuli.go
```diff
@@ -0,0 +1,136 @@
...
+	} else {
+		// Replace the series name with the value from the book
+		if id, ok := s.book[series]; ok {
+			buf = buf[:HeaderLength]
+			buf = append(buf, fmt.Sprintf(":%d", id)...)
+			binary.LittleEndian.PutUint16(buf[4:6], uint16(len(buf)))
+			binary.LittleEndian.PutUint16(buf[6:HeaderLength], uint16(0))
+			binary.LittleEndian.PutUint32(buf[:4], id)
+		} else {
+			panic("Unexpected series name")
```
RobAtticus: let's just return an error here? A long term goal I have is to reduce the amount of panics in the codebase, and given this func already returns errors, we should probably just do that

Lazin: Done

cmd/tsbs_generate_data/serialize/akumuli.go
Lazin: Done

RobAtticus:
Sorry it took so long, but merged!

PR #74, opened 2019-05-17
Make statProcessor an interface and create tests for BencRunner

blagojts:
The statProcessor responsible for gathering statistics
when executing queries was built as a struct. This commit
will change it to an interface to make the BenchmarkRunner code
more easier to test. This commit also adds some unit tests for
the benchmark runner that check if proper argument checks are
done, and if proper init happens when the Run method is called

### DIFF
--- a/query/benchmarker_test.go
+++ b/query/benchmarker_test.go
@@ -1,6 +1,9 @@
 package query
 
 import (
+	"io/ioutil"
+	"os"
+	"strings"
 	"sync"
 	"testing"
 )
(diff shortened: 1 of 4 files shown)


query/stat_processor.go
```diff
@@ -8,16 +8,40 @@ import (
...
-	prewarmQueries bool       // PrewarmQueries tells the StatProcessor whether we're running each query twice to prewarm the cache
-	c              chan *Stat // c is the channel for Stats to be sent for processing
-	limit          *uint64    // limit is the number of statistics to analyze before stopping
-	burnIn         uint64     // burnIn is the number of statistics to ignore before analyzing
-	printInterval  uint64     // printInterval is how often print intermediate stats (number of queries)
-	wg             sync.WaitGroup
+type statProcessor interface {
+	getArgs() *statProcessorArgs
+	sendStats(stats []*Stat)
+	sendStatsWarm(stats []*Stat)
```
antekresic: Since we are refactoring, maybe we should consider removing some of the stutter in the interface there. The word `Stats` seems superfluous in the function name.

`b.sp.sendStats(stats)` => `b.sp.send(stats)`

Same goes for `sendStats` and `sendStatsWarm`.

RobAtticus: I like this suggestion.
`sendStats` -> `send`
`sendStatsWarm` -> `sendWarm`

query/stat_processor.go
```diff
@@ -8,16 +8,40 @@ import (
...
+	process(workers uint)
+	CloseAndWait()
 }
 
-func (sp *statProcessor) sendStats(stats []*Stat) {
+type statProcessorArgs struct {
```
RobAtticus: good change 👍

query/benchmarker_test.go
```diff
@@ -52,3 +56,206 @@ func TestProcessorHandler(t *testing.T) {
...
+		},
+		onSend: func(_ []*Stat) {
+			lock.Lock()
+			sendStatsCalled = true
+			lock.Unlock()
+		},
+		wg: wg,
+	}
+
+	// query execution workers wait to receive a decoded query
```
RobAtticus: I prefer if comments are either complete sentences/paragraph with capitalization, etc, or a single line phrase.

query/benchmarker_test.go
```diff
@@ -52,3 +56,206 @@ func TestProcessorHandler(t *testing.T) {
...
+	_, err := os.Stat(dumbFileName)
+	if os.IsExist(err) {
+		t.Fatalf("File '%s' should not exist", dumbFileName)
+	}
+	b := &BenchmarkRunner{
+		fileName: dumbFileName,
+	}
+	defer func() {
+		if r := recover(); strings.HasPrefix(r.(string), "cannot open file for read") {
+			fmt.Println("Recovered in f", r)
```
RobAtticus: just to keep the output clean, can you make it only print if something goes wrong?

also most of these things should probably be lowercase to start, per Golang's error format

query/benchmarker_test.go
```diff
@@ -52,3 +56,206 @@ func TestProcessorHandler(t *testing.T) {
...
+	b.fileName = dumbFileName
+
+	// second call should use cached, not open another BuffReader
+	b.GetBufferedReader()
+}
+
+func TestBenchmarkRunnerRunPanicOnNoWorkers(t *testing.T) {
+	runner := &BenchmarkRunner{}
+	defer func() {
+		if r := recover(); r == "must have at least one worker" {
```
RobAtticus: same thing about the panics

PR #75, opened 2019-05-20
Combine both TimeInterval types into single type

RobAtticus:
Query generation and Cassandra's query running both used a type
called TimeInterval that did roughly the same thing. This change
combines the two into one type that can be used from the utils
package in internal/. This improves code reuse and keeps the two
representations in sync, and also increases the testability of
the code.


internal/utils/time_interval.go
```diff
@@ -0,0 +1,130 @@
+package utils
+
+import (
+	"fmt"
+	"math/rand"
+	"time"
+)
+
+const (
+	// ErrEndBeforeStart is an the error message for when a TimeInterval's end time
```
antekresic: Superfluous `the`? 

~Also, I'm from the camp which avoids over-commenting, which in this case, IMO, this would be. Sometimes unnecessary comments can be just as bad as missing comments. I do understand its a personal thing mostly.~

Scratch that, its exported. I'm blind.

internal/utils/time_interval.go
```diff
@@ -0,0 +1,130 @@
...
+type TimeInterval struct {
+	start time.Time
+	end   time.Time
+}
+
+// NewTimeInterval creates a new TimeInterval for a given start and end. If end
+// is a time.Time before start, then an error is returned.
+func NewTimeInterval(start, end time.Time) (*TimeInterval, error) {
+	if end.Before(start) {
+		return nil, fmt.Errorf(ErrEndBeforeStart)
```
antekresic: Readability nit: I would prefer to use `errors.New` here, even if it means a new import. `fmt.Errorf` uses that anyways and there is no formatting going on here.

internal/utils/time_interval.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Duration returns the time.Duration of the TimeInterval.
+func (ti *TimeInterval) Duration() time.Duration {
+	return ti.end.Sub(ti.start)
+}
+
+// Overlap detects whether the given TimeInterval overlaps with this
+// TimeInterval, assuming an inclusive start boundary and exclusive end
+// boundary.
+func (ti *TimeInterval) Overlap(other *TimeInterval) bool {
```
antekresic: Is this function missing a case? When the first interval starts before second one and ends after the second interval and vice-versa.

I feel this could be written simpler, exclude all the cases where there is no overlap first, since there are less of those. Everything else is overlap.

For example, keep the first check for opposite boundaries and do another check for start and end of this before start of that  and the last check for start and end of this after end of that.

internal/utils/time_interval.go
```diff
@@ -0,0 +1,130 @@
...
+	thisS := ti.Start()
```
antekresic: I feel `thisS` and `thatS` are not reader-friendly. Easily confused, hard to see which is which on a glance. Something like `firstS` and `secondS` would be much better IMO.

internal/utils/time_interval.go
antekresic: After reading the tests, I'm probably wrong on the first part of the comment. But I guess it just solidifies the second part and the comment under this one.

cmd/tsbs_generate_queries/databases/cassandra/devops.go
```diff
@@ -50,7 +50,10 @@ func (d *Devops) getHostWhere(nHosts int) []string {
 // AND time >= '$HOUR_START' AND time < '$HOUR_END'
 // GROUP BY minute ORDER BY minute ASC
 func (d *Devops) GroupByTime(qi query.Query, nHosts, numMetrics int, timeRange time.Duration) {
-	interval := d.Interval.RandWindow(timeRange)
+	interval, err := d.Interval.RandWindow(timeRange)
+	if err != nil {
+		panic(err.Error())
```
antekresic: We could wrap these errors for additional readability. Especially since we are panicking on them.

internal/utils/time_interval.go
RobAtticus: Hmm, I'm the opposite because I find often times I go back to error messages and add formatting options later. Even when I don't, I like the consistency of using or the other. I'm inclined to leave it like this until we do a more thorough errors refactoring to use best practices here:
https://dave.cheney.net/paste/gocon-spring-2016.pdf

Thoughts?

internal/utils/time_interval.go
RobAtticus: Fair point, I'll change to `s1`/`s2` and `e1`/`e2` since it's shorter

internal/utils/time_interval.go
antekresic: That sounds good!

internal/utils/time_interval.go
RobAtticus: The first part you mention is covered by the 3rd case -- first starts before the second, so we need to make sure the second one starts before the first one ends. Maybe it was a readability issue, so it's easier with clearer names.

But I like your solution better, it is cleaner/simpler. So I went with that.

cmd/tsbs_generate_queries/databases/cassandra/devops.go
RobAtticus: Happy to do that, thought this was a simpler change with less chance of me getting it wrong.

antekresic: Its not a big deal, we can leave it for later when they introduce error wrapping to `stdlib`.

RobAtticus:
Hey @antekresic , asking for a re-review because I added `MustRandWindow` for cleaner caller code. So this handles the `panic` issue by doing it in the library (we can eventually remove this in favor of bubbling up the errors).

I also rewrote the tests to use the `t.Run` trick I learned in your other PR :)

PR #76, opened 2019-05-21
Add unit tests to query generation functions for various databases

antekresic:
Covering query generation functions for Influx, ClickHouse and SiriDB
databases. Tests are covering basic pre-generated outputs and provide
visual sanity checks. More robust tests are left as a future task.

### DIFF
--- a/cmd/tsbs_generate_queries/databases/clickhouse/devops_test.go
+++ b/cmd/tsbs_generate_queries/databases/clickhouse/devops_test.go
@@ -1,9 +1,13 @@
 package clickhouse
 
 import (
+	"math/rand"
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/timescale/tsbs/cmd/tsbs_generate_queries/uses/devops"
+	"github.com/timescale/tsbs/query"
 )
 
 func TestDevopsGetHostWhereWithHostnames(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cmd/tsbs_generate_queries/databases/clickhouse/devops_test.go
```diff
@@ -84,3 +90,436 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+	expectedHumanLabel string
+	expectedHumanDesc  string
+	expectedQuery      string
+}
+
+func runTestCases(t *testing.T, testFunc func(*Devops, Case) query.Query, s time.Time, e time.Time, cases []Case) {
+	rand.Seed(123) // Setting seed for testing purposes.
+
+	for _, c := range cases {
+		t.Run(c.desc, func(t *testing.T) {
```
RobAtticus: wow, i never knew about this but it's exactly what i've wanted and (poorly) re-implemented all over the place.

cmd/tsbs_generate_queries/databases/influx/devops_test.go
```diff
@@ -41,6 +46,39 @@ func TestDevopsGetHostWhereWithHostnames(t *testing.T) {
...
+		},
+		{
+			desc:   "multi host (3)",
+			nHosts: 3,
+			want:   "(hostname = 'host_1' or hostname = 'host_8' or hostname = 'host_5')",
+		},
+	}
+
+	for _, c := range cases {
+		d := NewDevops(time.Now(), time.Now(), 10)
```
RobAtticus: you never use `desc` from `c`, so maybe wrap this in the `t.Run` call?

cmd/tsbs_generate_queries/databases/influx/devops_test.go
```diff
@@ -114,3 +360,122 @@ func TestDevopsFillInQuery(t *testing.T) {
 		t.Errorf("filled query has wrong path: got %s want /query?%s", got, encoded)
 	}
 }
+
+type Case struct {
```
RobAtticus: i know it ultimately doesn't matter, but maybe lowercase this so someone doesn't get confused about its exported-ness

cmd/tsbs_generate_queries/databases/influx/devops_test.go
```diff
@@ -114,3 +360,122 @@ func TestDevopsFillInQuery(t *testing.T) {
...
+	desc               string
+	input              int
+	fail               bool
+	failMsg            string
+	expectedHumanLabel string
+	expectedHumanDesc  string
+	expectedQuery      string
+}
+
+//func runTestCases(t *testing.T, testFunc func(*Devops, Case) query.Query, s time.Time, e time.Time, cases []Case) {
```
RobAtticus: remove?

cmd/tsbs_generate_queries/uses/devops/common.go
```diff
@@ -43,6 +43,18 @@ const (
 // for ease of testing
 var fatal = log.Fatalf
 
+// GetFatal allows us to get the current fatal logger function.
+// This is mostly to help with testing.
+func GetFatal() func(format string, v ...interface{}) {
```
RobAtticus: I understand why you needed this, but it seems a bit hacky to me, so I went and 'fixed' the underlying problem here:
https://github.com/timescale/tsbs/pull/77

It will also probably make the test methods above a bit more sane since you can just do a standard trick for capturing a panic

cmd/tsbs_generate_queries/databases/siridb/devops_test.go
```diff
@@ -94,3 +341,74 @@ func TestDevopsFillInQuery(t *testing.T) {
...
+
+			if c.fail {
+				f := devops.GetFatal()
+				devops.SetFatal(func(wg *sync.WaitGroup) func(string, ...interface{}) {
+					return func(s string, v ...interface{}) {
+						defer wg.Done()
+						if s != c.failMsg {
+							t.Fatalf("incorrect fail message: got %s, want %s", s, c.failMsg)
+						}
+						runtime.Goexit()
```
RobAtticus: moot point if we go with my other pr (#77), but is this necessary? we don't have to have the same semantics of log.Fatalf (i.e. killing the process). If we take this out, we can reduce the amount of defer/waitgroup also, making this an easier func to understand

cmd/tsbs_generate_queries/databases/influx/devops_test.go
antekresic: Ugh I missed that, will do!

cmd/tsbs_generate_queries/uses/devops/common.go
antekresic: I think I'll wait for that PR to be merged before changing this. Is that OK?

cmd/tsbs_generate_queries/databases/siridb/devops_test.go
antekresic: I agree but if we don't kill the process, fatal will be called multiple times in a test case and thats something that the existing function stops from happening. That means tests would fail and I would have to have multiple `fatalMsg`s to check each one.

I know it looks over-engineered but I didn't find a better way of replicating the current behavior in order that the tests reflect real-world usage.

cmd/tsbs_generate_queries/databases/influx/devops_test.go
antekresic: Make sense, I'm drawing a blank as to why I made it exported in the first place.

antekresic:
@RobAtticus updated the tests based on your feedback. Waiting with further changes until PR #77 is merged to master.

antekresic:
@RobAtticus I rebased the PR with your latest changes and fixed them up to reflect them.

Could you take another look at it?

Thanks.

cmd/tsbs_generate_queries/databases/clickhouse/devops_test.go
```diff
@@ -84,3 +88,443 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+
+	runTestCases(t, testFunc, start, end, cases)
+}
+
+func diffStrings(first, second string) int {
+	firstBytes := []byte(first)
+	secondBytes := []byte(second)
+
+	for i, byte := range firstBytes {
+		if byte != secondBytes[i] {
```
RobAtticus: small nit: `b` might be nicer so that its not shadowing the type `byte`

or you could just do `for i := range firstBytes` since you only care about the index

cmd/tsbs_generate_queries/databases/clickhouse/devops_test.go
```diff
@@ -84,3 +88,443 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
...
+						}
+
+						if r != c.failMsg {
+							t.Fatalf("incorrect fail message: got %s, want %s", r, c.failMsg)
+						}
+					}()
+
+					testFunc(d, c)
+				}()
+				return
```
RobAtticus: I think it's probably better to just wrap the other case in an `else` rather than early return. it's more obvious that they are two different paths that way

cmd/tsbs_generate_queries/databases/influx/devops_test.go
```diff
@@ -33,14 +36,51 @@ func TestDevopsGetHostWhereWithHostnames(t *testing.T) {
...
 	for _, c := range cases {
-		d := NewDevops(time.Now(), time.Now(), 10)
+		t.Run(c.desc, func(t *testing.T) {
+			d := NewDevops(time.Now(), time.Now(), 10)
 
-		if got := d.getHostWhereWithHostnames(c.hostnames); got != c.want {
-			t.Errorf("%s: incorrect output: got %s want %s", c.desc, got, c.want)
-		}
+			if got := d.getHostWhereWithHostnames(c.hostnames); got != c.want {
+				t.Errorf("%s: incorrect output: got %s want %s", c.desc, got, c.want)
```
RobAtticus: you can remove the `c.desc` arg from here now :)

cmd/tsbs_generate_queries/databases/influx/devops_test.go
```diff
@@ -69,12 +109,219 @@ func TestDevopsGetSelectClausesAggMetrics(t *testing.T) {
 	}
 
 	for _, c := range cases {
-		d := NewDevops(time.Now(), time.Now(), 10)
+		t.Run(c.desc, func(t *testing.T) {
+			d := NewDevops(time.Now(), time.Now(), 10)
+
+			if got := strings.Join(d.getSelectClausesAggMetrics(c.agg, c.metrics), ","); got != c.want {
+				t.Errorf("%s: incorrect output: got %s want %s", c.desc, got, c.want)
```
RobAtticus: remove `c.desc`

cmd/tsbs_generate_queries/databases/influx/devops_test.go
```diff
@@ -114,3 +361,77 @@ func TestDevopsFillInQuery(t *testing.T) {
...
+						}
+
+						if r != c.failMsg {
+							t.Fatalf("incorrect fail message: got %s, want %s", r, c.failMsg)
+						}
+					}()
+
+					testFunc(d, c)
+				}()
+				return
```
RobAtticus: same thing about the early return

cmd/tsbs_generate_queries/databases/siridb/devops_test.go
```diff
@@ -94,3 +339,65 @@ func TestDevopsFillInQuery(t *testing.T) {
...
+						}
+
+						if r != c.failMsg {
+							t.Fatalf("incorrect fail message: got %s, want %s", r, c.failMsg)
+						}
+					}()
+
+					testFunc(d, c)
+				}()
+				return
```
RobAtticus: same early return

cmd/tsbs_generate_queries/databases/clickhouse/devops_test.go
antekresic: That was supposed to be a short debug function, I'll remove it completely.

PR #77, opened 2019-05-21
Refactor generate_queries devops case to use errs

RobAtticus:
Previously the devops use case generation code used a call to
log.Fatalf when something went wrong. This makes it awkward to test
error conditions when generating queries from other packages, since
we need a way to (a) replace the unexported call to log.Fatalf and
(b) prevent the runtime from actually quitting.

It is better for the library to actually return errors on calls
that can fail, rather than either fataling or panicking. Now other
packages can handle the errors themselves and also test error
conditions in their packages as well.

This refactor was pruned a bit to bubble the 'panic' up one level
for now. When the actual generation code encounters the error
during normal execution, it will panic. But these are easier to
test for and don't require adding hooks to replace the 'fatal'
path in the original package.

### DIFF
--- a/cmd/tsbs_generate_queries/uses/devops/common.go
+++ b/cmd/tsbs_generate_queries/uses/devops/common.go
@@ -2,7 +2,6 @@ package devops
 
 import (
 	"fmt"
-	"log"
 	"math/rand"
 	"reflect"
 	"time"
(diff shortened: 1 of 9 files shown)


cmd/tsbs_generate_queries/uses/devops/common.go
```diff
@@ -40,9 +39,6 @@ const (
 	LabelHighCPU = "high-cpu"
 )
 
-// for ease of testing
```
antekresic: I must admit, I did find this ironic when I first read it 😄

RobAtticus: yeah enjoyed a nice chuckle as I removed it

PR #84, opened 2019-07-24
Improve Timescale query timing by confirming query completion

antekresic:
When using the `pgx` sql driver, running the query does not wait for a
response from the server. In order to verify that the query has returned
complete results, we must run `Rows.Next()` until it returns false
meaning we have fetched all the rows from the query result. Note that
this behavior is different than the current implementation of the `pq`
driver.


cmd/tsbs_run_queries_timescaledb/main.go
```diff
@@ -204,7 +204,13 @@ func (p *processor) ProcessQuery(q query.Query, isWarm bool) ([]*query.Stat, err
 	} else if p.opts.printResponse {
 		prettyPrintResponse(rows, tq)
 	}
+	// Fetching all the rows to confirm that the query is fully completed.
```
RobAtticus: we only need to do this in the case of using `pgx` right? Can we guard it with an if statement?

PR #89, opened 2019-09-02
Add new IoT query for Timescale and Influx databases

antekresic:
The new query fetches the last location of a single truck referenced by the
truck name. Truck name will be generated randomly from the scale which
represent the number of trucks currently available.

### DIFF
--- a/cmd/tsbs_generate_queries/databases/influx/iot_test.go
+++ b/cmd/tsbs_generate_queries/databases/influx/iot_test.go
@@ -1,16 +1,76 @@
 package influx
 
 import (
+	"fmt"
 	"math/rand"
+	"net/url"
 	"testing"
 	"time"
+
+	"github.com/timescale/tsbs/query"
+)
+
+const (
+	testScale = 10
 )
 
 type IoTTestCase struct {
 	desc               string
+	input              int
+	fail               bool
+	failMsg            string
 	expectedHumanLabel string
 	expectedHumanDesc  string
 	expectedQuery      string
+}
+
+func TestLastLocByTruck(t *testing.T) {
+	cases := []IoTTestCase{
+		{
+			desc:    "zero trucks",
+			input:   0,
+			fail:    true,
+			failMsg: "number of trucks cannot be < 1; got 0",
+		},
+		{
+			desc:    "more trucks than scale",
+			input:   2 * testScale,
+			fail:    true,
+			failMsg: "number of trucks (20) larger than total trucks. See --scale (10)",
+		},
+		{
+			desc:  "one truck",
+			input: 1,
+
+			expectedHumanLabel: "Influx last location by specific truck",
+			expectedHumanDesc:  "Influx last location by specific truck: random    1 trucks",
+			expectedQuery: `SELECT "name", "driver", "latitude", "longitude" 
+		FROM "readings" 
+		WHERE ("name" = 'truck_5') 
+		ORDER BY "time" 
+		LIMIT 1`,
+		},
+		{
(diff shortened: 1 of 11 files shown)


cmd/tsbs_generate_queries/databases/influx/iot_test.go
```diff
@@ -456,3 +512,43 @@ func TestTenMinutePeriods(t *testing.T) {
...
 
 }
+
+func runIoTTestCases(t *testing.T, testFunc func(*IoT, IoTTestCase) query.Query, s time.Time, e time.Time, cases []IoTTestCase) {
+	rand.Seed(123) // Setting seed for testing purposes.
+
+	for _, c := range cases {
+		t.Run(c.desc, func(t *testing.T) {
+			b := BaseGenerator{}
+			dq, err := b.NewIoT(s, e, 10)
```
blagojts: I saw a test above that was supposed to fail because the required number of trucks was 20, and this was greater than the scale. Can we extract this 10 to a const and use it in the test above also?

cmd/tsbs_generate_queries/databases/timescaledb/iot_test.go
```diff
@@ -6,17 +6,92 @@ import (
...
+	cases := []testCase{
+		{
+			desc:    "zero trucks",
+			input:   0,
+			fail:    true,
+			failMsg: "number of trucks cannot be < 1; got 0",
+		},
+		{
+			desc:    "more trucks than scale",
+			input:   20,
```
blagojts: this is the scale number i was referring to in my previous comment (also the one in the influx/iot_test)

cmd/tsbs_generate_queries/uses/common/common.go
```diff
@@ -32,3 +37,30 @@ func NewCore(start, end time.Time, scale int) (*Core, error) {
...
+// The subset of the permutation should have no duplicates and thus, can not be longer that original set
+// Ex.: 12, 7, 25 for numItems=3 and totalItems=30 (3 out of 30)
+func GetRandomSubsetPerm(numItems int, totalItems int) ([]int, error) {
+	if numItems > totalItems {
+		// Cannot make a subset longer than the original set
+		return nil, fmt.Errorf(errMoreItemsThanScale)
+	}
+
+	seen := map[int]bool{}
+	res := []int{}
```
blagojts: since we know the `numItems` we can use a fixed size array and so append doesn't allocate new memory

cmd/tsbs_generate_queries/uses/iot/common.go
```diff
@@ -68,11 +71,45 @@ func NewCore(start, end time.Time, scale int) (*Core, error) {
...
 }
 
+// GetRandomTrucks returns a random set of nTrucks from a given Core
+func (c *Core) GetRandomTrucks(nTrucks int) ([]string, error) {
+	return getRandomTrucks(nTrucks, c.Scale)
+}
+
+// getRandomTruckNames returns a subset of numTrucks names of a permutation of truck names,
+// numbered from 0 to totalTrucks.
+// Ex.: truck_12, truckhost_7, truck_25 for numTrucks=3 and totalTrucks=30 (3 out of 30)
```
blagojts: truckhost_7

cmd/tsbs_generate_queries/databases/influx/iot_test.go
RobAtticus: agree ^

RobAtticus:
LGTM

PR #92, opened 2019-09-18
Update float value precision for generated IoT use case data

antekresic:
Previously, floating point precision was unbounded. This commit
introduces some sensible precision limits to better emulate real world
data.

### DIFF
--- a/cmd/tsbs_generate_data/common/distribution_test.go
+++ b/cmd/tsbs_generate_data/common/distribution_test.go
@@ -1,0 +1,98 @@
+package common
+
+import (
+	"math"
+	"testing"
+)
+
+type mockDistribution struct {
+	AdvanceCalled bool
+	ReturnValue   float64
+}
+
+func (m *mockDistribution) Advance() {
+	m.AdvanceCalled = true
+}
+
+func (m *mockDistribution) Get() float64 {
+	return m.ReturnValue
+}
+
+func TestFloatPrecisionAdvance(t *testing.T) {
+
+	dist := &mockDistribution{}
+
+	fp := FP(dist, 1)
+
+	fp.Advance()
+
+	if !dist.AdvanceCalled {
+		t.Errorf("FloatPrecision Advance call did not call underlying distribution Advance method")
+	}
+}
+
+func TestFloatPrecisionGet(t *testing.T) {
+	testCases := []struct {
+		value   float64
+		results map[int]float64
+	}{
+		{
+			value: 1.234567890,
+			results: map[int]float64{
+				-1: 1,
+				0:  1,
+				1:  1.2,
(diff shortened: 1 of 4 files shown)


cmd/tsbs_generate_data/common/distribution_test.go
```diff
@@ -0,0 +1,70 @@
+package common
+
+import (
+	"math"
+	"testing"
+)
+
+const (
+	tolerance = 0.0000000001
```
RobAtticus: doesn't seem to be used?

cmd/tsbs_generate_data/common/distribution_test.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+}
+
+func TestFloatPrecisionGet(t *testing.T) {
+	testCases := []struct {
+		value   float64
+		results map[int]float64
+	}{
+		{
+			value: 1.234567890,
```
RobAtticus: can you add another case for an integer like 1.0? and just for fun, 0.0 as well? those values seem "special" enough to have their own tests

cmd/tsbs_generate_data/iot/diagnostics.go
```diff
@@ -24,16 +24,31 @@ var (
 
 	diagnosticsFields = []common.LabeledDistributionMaker{
 		{
-			Label:             labelFuelState,
-			DistributionMaker: func() common.Distribution { return &customFuelDistribution{common.CWD(fuelUD, 0, maxFuel, maxFuel)} },
+			Label: labelFuelState,
+			DistributionMaker: func() common.Distribution {
+				return common.FP(
+					&customFuelDistribution{common.CWD(fuelUD, 0, maxFuel, maxFuel)},
```
RobAtticus: when i last talked with @davidkohn88 he mentioned that this seemed to jump around a bit more erratically than one might expect. Does that seem to be a problem? It is a clamped random walk so I would imagine it wouldn't but haven't seen the output.

antekresic: I believe he was talking about current load which should be modeled differently (staying at one state for periods of time and changing linearly from state to state to simulate loading/unloading) but that would require some thought how to do it properly. 

Do you have any suggestions there?

RobAtticus: I'll have to think about it a bit, but it probably requires a different distributor. Maybe something like a ClampedRandomWalk that also takes a parameter for % chance that the value stays the same which we can set to like 80%.

RobAtticus: Can be a separate PR, let's merge this in for now.

PR #93, opened 2019-09-25
Reduce the variability in a Truck's Load in IoT

blagojts:
A truck's current load diagnostic was changing too much between measurements.
This change introduces a SaddledDistribution. A distribution that will change it's 
value only if a random value goes over a given threshold.

### DIFF
--- a/cmd/tsbs_generate_data/common/distribution.go
+++ b/cmd/tsbs_generate_data/common/distribution.go
@@ -206,3 +206,36 @@ func FP(step Distribution, precision int) *FloatPrecision {
 		precision: math.Pow(10, float64(precision)),
 	}
 }
+
+// LazyDistribution is a distribution that can change it's value
+// only if a "motivation" distribution provides a value above a specified threshold.
+// Otherwise it remains the same.
+type LazyDistribution struct {
+	motive    Distribution
+	step      Distribution
+	threshold float64
+}
+
+// LD returns a new LazyDistribution that returns a new value from "dist", if the "motavation" distribution,
+// fires above the threshold.
+func LD(motive, dist Distribution, threshold float64) *LazyDistribution {
+	return &LazyDistribution{
+		step:      dist,
+		motive:    motive,
+		threshold: threshold,
+	}
+}
+
+// Advance computes the next value of this distribution.
+func (d *LazyDistribution) Advance() {
+	d.motive.Advance()
+	if d.motive.Get() < d.threshold {
+		return
+	}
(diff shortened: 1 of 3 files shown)


cmd/tsbs_generate_data/common/distribution.go
```diff
@@ -206,3 +206,36 @@ func FP(step Distribution, precision int) *FloatPrecision {
...
+// SaddledDistribution is a distribution that can change it's value
+// only if a "saddle" distribution provides a value above a specified threshold.
+// Otherwise it remains the same.
+type SaddledDistribution struct {
+	Saddle    Distribution
+	Step      Distribution
+	Threshold float64
+}
+
+// SCD returns a new SaddledClampedDistribution based on a given saddle binary distribution and a uniform distribution.
```
RobAtticus: `SD`

cmd/tsbs_generate_data/common/distribution_test.go
```diff
@@ -96,3 +96,50 @@ func TestFloatPrecisionGet(t *testing.T) {
...
+		saddle    Distribution
+		step      Distribution
+		expect    float64
+	}{
+		{
+			desc:      "threshold not satisfied, value remains the same",
+			threshold: 1,
+			saddle:    &mockDistribution{ReturnValue: 0},
+			step:      CWD(&mockDistribution{ReturnValue: 1}, 2, 4, 2),
+			expect:    float64(2),
```
RobAtticus: don't think you need to cast this, I think the compiler infers it given the type of the field

cmd/tsbs_generate_data/common/distribution.go
```diff
@@ -206,3 +206,36 @@ func FP(step Distribution, precision int) *FloatPrecision {
 		precision: math.Pow(10, float64(precision)),
 	}
 }
+
+
+// SaddledDistribution is a distribution that can change it's value
+// only if a "saddle" distribution provides a value above a specified threshold.
+// Otherwise it remains the same.
+type SaddledDistribution struct {
+	Saddle    Distribution
```
RobAtticus: I think all of these can be unexported. Don't want somebody accidentally messing with struct fields

blagojts:
> Looks good to me -- saddle distribution didn't immediately come up on googling, where did the name come from?

It's something we used to use in college when describing how a search algorithm should behave, when it would find a local min, the local minimum would look like a saddle (a drop in a line) and the algorithm would need a bit of extra strength to get out of the saddle

PR #94, opened 2019-09-30
Fix the way to set flags in scripts due to switching to viper lib

hagen1778:
Since flag parse lib was changed to viper - old way for passing flags in scripts doesn't work:
```
FORMATS=timescaledb ./scripts/generate_data.sh
Generating data_timescaledb_cpu-only_4000_2019-09-01T00:00:00Z_2019-09-03T00:00:00Z_10s_123.dat.gz:
unknown shorthand flag: 'f' in -format
```

 This PR supposed to set the same way for passing flags for all scripts.


bboule:
@blagojts when you have a second could you take a look.

RobAtticus:
Thanks for this, sorry for the delay!

PR #96, opened 2019-10-04
Add VictoriaMetrics support

hagen1778:
This PR adds support for opensource timeseries database [VictoriaMetrics](https://github.com/VictoriaMetrics/VictoriaMetrics).

VM supports InfluxDB ingestion protocol, so `tsbs_load_victoriametrics` is very similar to `tsbs_load_influx`.

VictoriaMetrics supports [PromQL](https://prometheus.io/docs/prometheus/latest/querying/basics/) which has some limitations comparing to SQL. Because of this VictoriaMetrics query generator lacks for implementation of query types `groupby-orderby-limit` and `lastpoint` for `devops` use-case and all queries for `iot` use-case. But for other cases it still may be comparable with other databases.

We also believe this PR may help with adding of [Prometheus](https://github.com/prometheus/prometheus) to the list of supported databases.


RobAtticus:
Hi, thanks for this, we'll put it on our radar to review!

tenmozes:
@RobAtticus could you please take a look?

RobAtticus:
I've been out at conferences, so I might be able to next week. @blagojts if you have some spare cycles, make take an initial pass?

syepes:
+1

hagen1778:
Hi @RobAtticus! Any chance to get it reviewed?

tenmozes:
Hi folks, any updates on this PR?

hagen1778:
Hello! Any updates on this?

freeseacher:
Hi! any news ?

cmd/tsbs_generate_queries/databases/victoriametrics/devops.go
```diff
@@ -0,0 +1,161 @@
...
+// HighCPUForHosts populates a query that gets CPU metrics when the CPU has high
+// usage between a time period for a number of hosts (if 0, it will search all hosts),
+// e.g. in pseudo-PromQL:
+//
+// max(
+//	max_over_time(
+// 		{__name__=~'cpu_.*', hostname=~"hostname1|hostname2...|hostnameN"}[12h]
+// 	)
+// ) by (hostname) > 90
+func (d *Devops) HighCPUForHosts(qq query.Query, nHosts int) {
```
toddlipcon: This seems to be returning very different data than infuxdb -- the 12h grouping here isn't present in the equivalent influx or timescale query. Thus those systems return several hundred MB of JSON whereas Victoria is returning only two data points per host, making it way faster.

hagen1778: Thanks for noticing! I removed this case since it is unclear how to support it via MetricsQL/PromQL. The query model assumes step (grouping by time) to be always set. So to satisfy this case we need to pass `log-interval` flag value used for generating data which we don't know when generating queries. I hope other cases are consistent and comparable with other databases. Thank you again!

README.md
```diff
@@ -75,8 +76,10 @@ cases are implemented for each database:
 |MongoDB|X|
 |SiriDB|X|
 |TimescaleDB|X|X|
+|VictoriaMetrics|X²||
 
 ¹ Does not support the `groupby-orderby-limit` query
+² Does not support the `groupby-orderby-limit` and `lastpoint` queries
```
RobAtticus: needs to include that it does not support `high-cpu-1` and `high-cpu-all`

cmd/tsbs_run_queries_victoriametrics/main.go
```diff
@@ -0,0 +1,127 @@
+// tsbs_run_queries_clickhouse speed tests ClickHouse using requests from stdin or file.
```
RobAtticus: comment needs to be updated

docs/victoriametrics.md
```diff
@@ -0,0 +1,108 @@
...
+---
+
+## Generating queries
+
+VictoriaMetrics query language MetricsQL has some limitations comparing to SQL.
+Because of this VictoriaMetrics query generator lacks for implementation of query 
+types for `devops` use-case: 
+* `groupby-orderby-limit` - results are always ordered by time and can't be limited;
+* `lastpoint` - can't be queried if datapoint is older than 5 minutes; 
+* `high-cpu` - can't be queried without grouping by step.
```
RobAtticus: `high-cpu-1`/`high-cpu-all`

hagen1778:
Thx for review @RobAtticus! Addressed the comments. Pls let me know if there is anything else I can do.

hagen1778:
@RobAtticus  ping

AlekseyKanaev:
Any updates on the PR?

RobAtticus:
Merged! Thanks again for your patience

PR #99, opened 2019-10-15
Add ability to build Docker image

erimatnor:
This change adds a Dockerfile to build a Docker image for TSBS. This
makes it easier for people to quickly test TSBS and run benchmarks on, e.g., Kubernetes.

This also bumps Go modules to depend on Go 1.13.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+ARG GO_VERSION=1.13.1
+FROM golang:${GO_VERSION}-alpine
+
+ARG TSBS_VERSION=master
+RUN apk update && apk add --no-cache --virtual .build-deps git \
+    && mkdir -p ${GOPATH}/src/github.com/timescale/ \
+    && cd ${GOPATH}/src/github.com/timescale/ \
+    && git clone --depth=1 --branch ${TSBS_VERSION} https://github.com/timescale/tsbs.git \
+    && cd $GOPATH/src/github.com/timescale/tsbs \
+    && go get -d -v ./... \
+    && go build -o /go/bin/ ./... \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG GO_VERSION=1.13.1
+FROM golang:${GO_VERSION}-alpine
+
+ARG TSBS_VERSION=master
+RUN apk update && apk add --no-cache --virtual .build-deps git \
+    && mkdir -p ${GOPATH}/src/github.com/timescale/ \
```
blagojts: we're using go modules, no need to clone into $GOPATH and to do go get
so we can remove two lines (mkdir line 6, and go get line 10)

blagojts: also, I would suggest using multi-stage build (https://docs.docker.com/develop/develop-images/multistage-build/), so you don't need to worry about deleting the source and apk del. You can just copy the built binary

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/timescale/tsbs
 
-go 1.12
+go 1.13
```
RobAtticus: I don't think is actually a good change (we might want to drop it to `1.11` even). This is supposed to be the minimum version of Go that this module works with, and afaik there isn't any reason people with `1.12` shouldn't be able to build it. That is, we aren't depending on any `1.13` specific APIs

Dockerfile
RobAtticus: Yeah I agree with all of this.

And now `1.13.3` is the latest

blagojts:
Closing due of staleness

PR #101, opened 2019-10-24
enable full percentile spectrum analysis on latency. enable maintaining fixed rps during query benchmarks

filipecosta90:
# Full percentile spectrum analysis
This PR enable full percentile spectrum analysis on query latency so that we can fully understand the performance and stability characteristics of the system we are measuring. 

Right now tsbs is displaying single values targets like the quantile 50, average, min and max. 
Considering that latency does not obey to a normal distribution we should enable ourselfs to do full percentile spectrum analysis. Just looking at a number might be misleading. 

A useful visualization for pointing out the folly of chasing a mean measurement is illustrated below - they have nearly identical simple descriptive statistics, yet have very different distributions and appear very different when graphed.

image

Source: https://en.wikipedia.org/wiki/Anscombe%27s_quartet

_______
# Sustainable Throughput:
With that in mind, and to really understand a system behavior we also cant relly solely on doing the full percentile analysis while stressing the system to it's maximum RPS. We need to be able to compare the behavior under different throughputs and/or configurations, to be able to get the best "Sustainable Throughput: The throughput achieved while safely maintaining service levels. 

Adding a further reading slide that will help you understand why this is so important:

https://www.azul.com/files/HowNotToMeasureLatency_LLSummit_NYC_12Nov2013.pdf

image

_______
# Enabling full percentile spectrum and Sustainable Throughput analysis on tsbs:

The previous features can be applied with the changes added on this PR using the following flags on the `tsbs_run_queries_*` binaries:
-` --hdr-latencies` : enable writing the High Dynamic Range (HDR) Histogram of Response Latencies to the file with the name specified by this. By default no file will be saved. 
-` --max-rps` : enable limiting the rate of queries per second, 0 = no limit. By default no limit is specified and the binaries will behave in the same manner as before. 

## quick example of the beneficts of this addition
Take into consideration that the presented values were obtained to illustrate the added features and not to compared the used systems. This was a local test that should not be used for any other thing rather than to exemplify the new way of assessing stability and performance across the entire latency spectrum when using HDR histograms and rate-limiting.  

### Step 1 - Setup
Running two example docker containers from clickhouse and  timescale in the following manner:
```
docker run -d -p 9000:9000 --ulimit nofile=262144:262144 yandex/clickhouse-server
docker run -d -e POSTGRES_PASSWORD=password -p 5432:5432 timescale/timescaledb
```
### Step 2 - Retrieve sample values
Running two example docker containers from clickhouse and  timescale in the following manner:
```
MAX_QUERIES=1000 MAX_RPS=100 ./scripts/full_cycle_minitest_clickhouse.sh
MAX_QUERIES=1000 MAX_RPS=100 PASSWORD="password" ./scripts/full_cycle_minitest_timescaledb.sh
```
Sample output ( please denote the added `Interval query rate` and `Overall query rate` values
```
(...)
After 100 queries with 1 workers:
Interval query rate: 99.74 queries/sec  Overall query rate: 99.74 queries/sec
ClickHouse CPU over threshold, 1 host(s):
min:     6.03ms, med:     7.36ms, mean:     7.59ms, max:   11.82ms, stddev:     1.12ms, sum:   0.8sec, count: 100
all queries                             :
min:     6.03ms, med:     7.36ms, mean:     7.59ms, max:   11.82ms, stddev:     1.12ms, sum:   0.8sec, count: 100
(...)
```
Final Query benchmark output ( please denote the info regarding in which file the HDR latency file was saved 
```
(...)
Run complete after 1000 queries with 1 workers (Overall query rate 98.42 queries/sec):
ClickHouse CPU over threshold, 1 host(s):
min:     5.78ms, med:     7.52ms, mean:     7.90ms, max:  111.44ms, stddev:     3.54ms, sum:   7.9sec, count: 1000
all queries                             :
min:     5.78ms, med:     7.52ms, mean:     7.90ms, max:  111.44ms, stddev:     3.54ms, sum:   7.9sec, count: 1000
Saving High Dynamic Range (HDR) Histogram of Response Latencies to 100rps_clickhouse_query_high-cpu-1.hdr
wall clock time: 10.178806sec
```
### Step 3 - Use HdrHistogram Plotter
Using the two first output files from the simple full cycle mini tests we can plot the Latency by Percentile Distribution using http://hdrhistogram.github.io/HdrHistogram/plotFiles.html

- [100rps_clickhouse_query_high-cpu-1.txt](https://github.com/filipecosta90/tsbs/files/3767556/100rps_clickhouse_query_high-cpu-1.txt)
- [100rps_timescaledb_query_high-cpu-1.txt](https://github.com/filipecosta90/tsbs/files/3767558/100rps_timescaledb_query_high-cpu-1.txt)
- [100rps_clickhouse_query_cpu-max-all-1.txt](https://github.com/filipecosta90/tsbs/files/3767555/100rps_clickhouse_query_cpu-max-all-1.txt)
- [100rps_timescaledb_query_cpu-max-all-1.txt](https://github.com/filipecosta90/tsbs/files/3767557/100rps_timescaledb_query_cpu-max-all-1.txt)

Which will enable us to get the "full picture" of latency distribution. If we only analyzed the P99 latency we would be tempted to state that clickhouse had better values. However, if you look and the second image you see that on overall  the stability of timescale is greater. Take into consideration that obtained values are only for illustration purposes and to justify that single values targets can lead us to the wrong conclusions. 

Screen Shot 2019-10-24 at 1 46 58 PM
Screen Shot 2019-10-24 at 1 46 44 PM

# credits 
HdrHistogram was originally authored by Gil Tene (@giltene) and placed in the public domain, as explained at http://creativecommons.org/publicdomain/zero/1.0/, with ports by Mike Barker (@mikeb01) (C), Matt Warren (@mattwarren) (C#), Darach Ennis (@darach) (Erlang), Alec Hothan (@ahothan) (Python), Coda Hale (@codahale) (Go) (which i needed to revamp), Jon Gjengset (@jonhoo) (Rust), and Alexandre Victoor (@alexvictoor) (JavaScript)

### DIFF
--- a/query/benchmarker.go
+++ b/query/benchmarker.go
@@ -4,12 +4,14 @@ import (
 	"bufio"
 	"fmt"
 	"log"
+	"math"
 	"os"
 	"runtime/pprof"
 	"sync"
 	"time"
 
 	"github.com/spf13/pflag"
+	"golang.org/x/time/rate"
 )
 
 const (
@@ -24,7 +26,9 @@ const (
 type BenchmarkRunnerConfig struct {
 	DBName         string `mapstructure:"db-name"`
 	Limit          uint64 `mapstructure:"max-queries"`
+	LimitRps       uint64 `mapstructure:"max-rps"`
 	MemProfile     string `mapstructure:"memprofile"`
+	HDRLatenciesFile     string `mapstructure:"hdr-latencies"`
 	Workers        uint   `mapstructure:"workers"`
 	PrintResponses bool   `mapstructure:"print-responses"`
 	Debug          int    `mapstructure:"debug"`
@@ -39,8 +43,10 @@ func (c BenchmarkRunnerConfig) AddToFlagSet(fs *pflag.FlagSet) {
 	fs.String("db-name", "benchmark", "Name of database to use for queries")
 	fs.Uint64("burn-in", 0, "Number of queries to ignore before collecting statistics.")
 	fs.Uint64("max-queries", 0, "Limit the number of queries to send, 0 = no limit")
+	fs.Uint64("max-rps", 0, "Limit the rate of queries per second, 0 = no limit")
 	fs.Uint64("print-interval", 100, "Print timing stats to stderr after this many queries (0 to disable)")
 	fs.String("memprofile", "", "Write a memory profile to this file.")
+	fs.String("hdr-latencies", "", "Write the High Dynamic Range (HDR) Histogram of Response Latencies to this file.")
 	fs.Uint("workers", 1, "Number of concurrent requests to make.")
 	fs.Bool("prewarm-queries", false, "Run each query twice in a row so the warm query is guaranteed to be a cache hit")
 	fs.Bool("print-responses", false, "Pretty print response bodies for correctness checking (default false).")
@@ -70,7 +76,7 @@ func NewBenchmarkRunner(config BenchmarkRunnerConfig) *BenchmarkRunner {
 		burnIn:         runner.BurnIn,
 	}
 
-	runner.sp = newStatProcessor(spArgs)
+	runner.sp = newStatProcessor(spArgs, runner.HDRLatenciesFile)
 	return runner
 }
 
@@ -139,13 +145,22 @@ func (b *BenchmarkRunner) Run(queryPool *sync.Pool, processorCreateFn ProcessorC
 	b.ch = make(chan Query, b.Workers)
 
 	// Launch the stats processor:
-	go b.sp.process(b.Workers)
+	go b.sp.process(b.Workers,b.HDRLatenciesFile)
+
+	var requestRate = rate.Limit(math.MaxFloat64)
+	var requestBurst = 0
+	if b.LimitRps != 0 {
+		requestRate = rate.Limit(b.LimitRps)
+		requestBurst = int(b.Workers)
+	}
+
+	var rateLimiter *rate.Limiter = rate.NewLimiter(requestRate, requestBurst)
 
 	// Launch query processors
 	var wg sync.WaitGroup
 	for i := 0; i < int(b.Workers); i++ {
 		wg.Add(1)
-		go b.processorHandler(&wg, queryPool, processorCreateFn(), i)
+		go b.processorHandler(&wg, rateLimiter, queryPool, processorCreateFn(), i)
 	}
 
 	// Read in jobs, closing the job channel when done:
@@ -177,9 +192,11 @@ func (b *BenchmarkRunner) Run(queryPool *sync.Pool, processorCreateFn ProcessorC
 	}
 }
 
-func (b *BenchmarkRunner) processorHandler(wg *sync.WaitGroup, queryPool *sync.Pool, processor Processor, workerNum int) {
+func (b *BenchmarkRunner) processorHandler(wg *sync.WaitGroup, rateLimiter *rate.Limiter, queryPool *sync.Pool, processor Processor, workerNum int) {
 	processor.Init(workerNum)
 	for query := range b.ch {
+		for rateLimiter.Allow() == false {
+		}
 		stats, err := processor.ProcessQuery(query, false)
 		if err != nil {
 			panic(err)
--- a/query/benchmarker.go
+++ b/query/benchmarker.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"fmt"
 	"log"
-	"math"
 	"os"
 	"runtime/pprof"
 	"sync"
@@ -26,7 +25,7 @@ const (
 type BenchmarkRunnerConfig struct {
 	DBName           string `mapstructure:"db-name"`
 	Limit            uint64 `mapstructure:"max-queries"`
-	LimitRps         uint64 `mapstructure:"max-rps"`
+	LimitRPS         uint64 `mapstructure:"max-rps"`
 	MemProfile       string `mapstructure:"memprofile"`
 	HDRLatenciesFile string `mapstructure:"hdr-latencies"`
 	Workers          uint   `mapstructure:"workers"`
@@ -148,14 +147,7 @@ func (b *BenchmarkRunner) Run(queryPool *sync.Pool, processorCreateFn ProcessorC
 	// Launch the stats processor:
 	go b.sp.process(b.Workers)
 
-	var requestRate = rate.Limit(math.MaxFloat64)
-	var requestBurst = 0
-	if b.LimitRps != 0 {
-		requestRate = rate.Limit(b.LimitRps)
-		requestBurst = int(b.Workers)
-	}
-
-	var rateLimiter *rate.Limiter = rate.NewLimiter(requestRate, requestBurst)
+	rateLimiter := getRateLimiter(b.LimitRPS,b.Workers)
 
 	// Launch query processors
 	var wg sync.WaitGroup
@@ -196,7 +188,7 @@ func (b *BenchmarkRunner) Run(queryPool *sync.Pool, processorCreateFn ProcessorC
 func (b *BenchmarkRunner) processorHandler(wg *sync.WaitGroup, rateLimiter *rate.Limiter, queryPool *sync.Pool, processor Processor, workerNum int) {
 	processor.Init(workerNum)
 	for query := range b.ch {
-		r := rateLimiter.ReserveN(time.Now(), 1)
+		r := rateLimiter.Reserve()
 		time.Sleep(r.Delay())
 
 		stats, err := processor.ProcessQuery(query, false)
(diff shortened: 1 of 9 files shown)


RobAtticus:
At first glance this PR looks awesome @filipecosta90 ! I assigned myself and two others to help review it, but totally agree with the thoughts you put forward about being able to analyze the whole sequence of values. Thanks for doing this!

scripts/full_cycle_minitest_clickhouse.sh
```diff
@@ -1,15 +1,19 @@
 #!/bin/bash
```
blagojts: maybe a comment at the top to explain what the script does

query/stat_processor.go
```diff
@@ -64,7 +73,7 @@ func (sp *defaultStatProcessor) sendWarm(stats []*Stat) {
 
 // process collects latency results, aggregating them into summary
 // statistics. Optionally, they are printed to stderr at regular intervals.
-func (sp *defaultStatProcessor) process(workers uint) {
+func (sp *defaultStatProcessor) process(workers uint,latencyFile string ) {
```
blagojts: the latency file is passed `newStatProcessor` no need for it here

query/stats.go
```diff
@@ -57,53 +57,41 @@ func (s *Stat) reset() *Stat {
 
 // statGroup collects simple streaming statistics.
 type statGroup struct {
-	min    float64
-	max    float64
-	mean   float64
+	latencyHDRHistogram *hdrhistogram.Histogram
 	sum    float64
 	values []float64
```
blagojts: values is kept, but other than the initial/first recorded value, the others are not appended to the array

filipecosta90:
> Looks ok, I have just a few small requests for comments and a bit of a reorg for the latencyFile arg

@blagojts can you please revise the latest changes

filipecosta90:
@RobAtticus considering @blagojts approved it, does it need further changes or are we good to go?

RobAtticus:
@filipecosta90 Let me take a closer look tomorrow but I think we should be pretty close to getting this in.

One higher level question is what you think the value is of using the HDR histogram library vs just an output as a CSV of the latency measurements that are seen?

query/stats.go
```diff
@@ -57,97 +57,68 @@ func (s *Stat) reset() *Stat {
...
-	s.m += (n - oldM) / float64(s.count)
-	s.s += (n - oldM) * (n - s.m)
-	s.stdDev = math.Sqrt(s.s / (float64(s.count) - 1.0))
 }
 
 // string makes a simple description of a statGroup.
 func (s *statGroup) string() string {
-	return fmt.Sprintf("min: %8.2fms, med: %8.2fms, mean: %8.2fms, max: %7.2fms, stddev: %8.2fms, sum: %5.1fsec, count: %d", s.min, s.median(), s.mean, s.max, s.stdDev, s.sum/1e3, s.count)
+	return fmt.Sprintf("min: %8.2fms, med: %8.2fms, mean: %8.2fms, max: %7.2fms, stddev: %8.2fms, sum: %5.1fsec, count: %d",
+		float64(s.latencyHDRHistogram.Min())/10e2,
```
RobAtticus: can probably just use `s.Min()`, etc for these so that the calculations don't get out of sync

filipecosta90:
> @filipecosta90 Let me take a closer look tomorrow but I think we should be pretty close to getting this in.
> 
> One higher level question is what you think the value is of using the HDR histogram library vs just an output as a CSV of the latency measurements that are seen?

I think we can benefict from the visualization tools that are in-placed [see example](https://hdrhistogram.github.io/HdrHistogramJSDemo/plotFiles.html) and save the histograms in a lossless format that will enable us to merge several ones, keep them for later, compare them on a common tool, etc...

query/stats.go
filipecosta90: resolved in the new commit

query/benchmarker.go
```diff
@@ -22,25 +24,29 @@ const (
...
-	Workers        uint   `mapstructure:"workers"`
-	PrintResponses bool   `mapstructure:"print-responses"`
-	Debug          int    `mapstructure:"debug"`
-	FileName       string `mapstructure:"file"`
-	BurnIn         uint64 `mapstructure:"burn-in"`
-	PrintInterval  uint64 `mapstructure:"print-interval"`
-	PrewarmQueries bool   `mapstructure:"prewarm-queries"`
+	DBName           string `mapstructure:"db-name"`
+	Limit            uint64 `mapstructure:"max-queries"`
+	LimitRps         uint64 `mapstructure:"max-rps"`
```
RobAtticus: not: go style suggests this should be `LimitRPS`

query/benchmarker.go
```diff
@@ -141,11 +148,20 @@ func (b *BenchmarkRunner) Run(queryPool *sync.Pool, processorCreateFn ProcessorC
 	// Launch the stats processor:
 	go b.sp.process(b.Workers)
 
+	var requestRate = rate.Limit(math.MaxFloat64)
```
RobAtticus: can just use `rate.Inf` which is the same thing

query/benchmarker.go
```diff
@@ -177,9 +193,12 @@ func (b *BenchmarkRunner) Run(queryPool *sync.Pool, processorCreateFn ProcessorC
 	}
 }
 
-func (b *BenchmarkRunner) processorHandler(wg *sync.WaitGroup, queryPool *sync.Pool, processor Processor, workerNum int) {
+func (b *BenchmarkRunner) processorHandler(wg *sync.WaitGroup, rateLimiter *rate.Limiter, queryPool *sync.Pool, processor Processor, workerNum int) {
 	processor.Init(workerNum)
 	for query := range b.ch {
+		r := rateLimiter.ReserveN(time.Now(), 1)
```
RobAtticus: `rateLimited.Reserve` is shorthand for this

query/benchmarker_test.go
```diff
@@ -36,8 +38,12 @@ func TestProcessorHandler(t *testing.T) {
 	var wg sync.WaitGroup
 	qPool := &testQueryPool
 	wg.Add(2)
-	go b.processorHandler(&wg, qPool, p1, 0)
-	go b.processorHandler(&wg, qPool, p2, 5)
+	var requestRate = rate.Limit(math.MaxFloat64)
```
RobAtticus: maybe a helper function called `getInfLimiter` since this is used in a couple of test functions

query/benchmarker.go
RobAtticus: to make this more testable, we can add an unexported method for getting the rateLimiter based on LimitRPS and Workers:
```go
func getRateLimiter(limitRPS uint64, workers uint) *rate.Limiter
```

Then it can be tested independently to make sure it returns the right things when limit is 0 or not.

query/stat_processor.go
```diff
@@ -131,18 +152,33 @@ func (sp *defaultStatProcessor) process(workers uint) {
...
 	err = writeStatGroupMap(os.Stdout, statMapping)
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	if len(sp.args.hdrLatenciesFile) > 0  {
+		_, _ = fmt.Printf("Saving High Dynamic Range (HDR) Histogram of Response Latencies to %s\n", sp.args.hdrLatenciesFile)
+
+		d1 := []byte(statMapping[allQueriesLabel].latencyHDRHistogram.PercentilesPrint(10, 1000.0))
+		fErr := ioutil.WriteFile(sp.args.hdrLatenciesFile, d1, 0644)
```
RobAtticus: i think you're ok reusing `err` here

query/stats.go
```diff
@@ -57,97 +57,76 @@ func (s *Stat) reset() *Stat {
...
-	m      float64
-	s      float64
-	stdDev float64
-
 	count int64
 }
 
 // newStatGroup returns a new StatGroup with an initial size
 func newStatGroup(size uint64) *statGroup {
+	// This latency Histogram could be used to track and analyze the counts of
```
RobAtticus: What happens for queries that take longer than 30s?

query/stats.go
```diff
@@ -57,97 +57,76 @@ func (s *Stat) reset() *Stat {
...
-	}
-
-	if n < s.min {
-		s.min = n
-	}
-	if n > s.max {
-		s.max = n
-	}
-
+	s.latencyHDRHistogram.RecordValue(int64(n * 10e2))
```
RobAtticus: nit: can this be written as `1e3`, the 10 throws me off since it makes the exponent off by 1 from what i'm used to.

Also since its used so often, probably should be a `const`, like `precisionFactor`?

RobAtticus:
Sorry a few more small things/nits, otherwise looks good!

query/stats.go
filipecosta90: it will return an error "value %d is too large to be recorded". will increase the limit to 60min since the overhead is not that much.

query/benchmarker_test.go
```diff
@@ -323,3 +336,27 @@ func (mp *mockProcessor) Init(workerNum int) { mp.initCalled = true }
 func (mp *mockProcessor) ProcessQuery(q Query, isWarm bool) ([]*Stat, error) {
 	return mp.processRes, mp.processErr
 }
+
+func Test_getRateLimiter(t *testing.T) {
```
RobAtticus: one last nit: `TestGetRateLimiter` to match format we already have

filipecosta90: solved in this new commit. thanks for the fast iterations over the changes :)

PR #111, opened 2019-12-30
Add use compatibility table to README

RobAtticus:
As the number of use cases and databases increase, it is useful for
users to discern which databases currently support which use cases.
This PR adds a table to the use case section with this information
and new databases should keep the table up-to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,36 +6,38 @@ at https://github.com/influxdata/influxdb-comparisons.
 
 Current databases supported:
 
-+ TimescaleDB [(supplemental docs)](docs/timescaledb.md)
-+ MongoDB [(supplemental docs)](docs/mongo.md)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,56 +6,78 @@ at https://github.com/influxdata/influxdb-comparisons.
 
 Current databases supported:
 
-+ TimescaleDB [(supplemental docs)](docs/timescaledb.md)
-+ MongoDB [(supplemental docs)](docs/mongo.md)
-+ InfluxDB [(supplemental docs)](docs/influx.md)
++ Akumuli [(supplemental docs)](docs/akumuli.md)
```
RobAtticus: Since the list is pretty lengthy now, it should probably be in ABC order

PR #123, opened 2020-05-07
port config has been added for clickhouse

AlekseyKanaev:
There may be situations when standard clickhouse port is not enough.

### DIFF
--- a/cmd/tsbs_load_clickhouse/main.go
+++ b/cmd/tsbs_load_clickhouse/main.go
@@ -23,6 +23,7 @@ const (
 // Program option vars:
 var (
 	host     string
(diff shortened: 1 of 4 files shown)


cmd/tsbs_load_clickhouse/main.go
```diff
@@ -23,6 +23,7 @@ const (
 // Program option vars:
 var (
 	host     string
+	port     string
```
RobAtticus, line 26: this seems like it should be an `int32` or `int64` and not a string? seems too risky to allow a string where somebody could inject all sorts of weirdness into the constructed connection URIs

PR #128, opened 2020-06-19
Adds Example for timescaledb Data Load Benchmark

tylerfontaine:
This adds a concrete example for running data load benchmarks against a 
remote timescaledb server, calling the binary directly due to some 
assumptions made by the load scripts which made this difficult 
previously.


RobAtticus (approved):
Sorry for the delay on this

PR #135, opened 2020-10-02
Reference TimescaleDB documentation

wilfriedroset:
While having tsbs to generate data is useful but it can be time saver to
redirect users to the documentation if they are looking for datasets to
bench their timescale.


blagojts (commented):
The mentioned data sets are not in a TSBS suitable format, so they are not a good fit to be mentioned in this repository

PR #148, opened 2021-01-13
Add the --analyze option for tsbs_load_timescaledb

mdcallag:
The default for the option is false and vacuum analyze is done for each table when true. Even for
insert-only workloads Postgres vacuum is useful as it sets bits in the visibility map so that secondary
index scans can really be index-only.

Note that DbCreatorCloser.Close used to be called after the load and was changed as part of the
refactoring to get called when Loader.preRun finishes (which is after create table, but before the
load is started). I assume that change was a mistake and restore the old behavior here.

Example output is below. The output from Close() is interleaved with the stats reporter output
because the stats reporter goroutine isn't stopped before process exist and a pending PR fixes that.
---
time,per. metric/s,metric total,overall metric/s,per. row/s,row total,overall row/s

Summary:
loaded 17452800 metrics in 9.803sec with 4 workers (mean rate 1780368.33 metrics/sec)
loaded 1555200 rows in 9.803sec with 4 workers (mean rate 158646.68 rows/sec)
Close: vacuum analyze tags
Close: vacuum analyze cpu
Close: vacuum analyze diskio
1610580022,1745255.84,1.745280E+07,1745255.84,155517.85,1.555200E+06,155517.85
Close: vacuum analyze nginx
Close: vacuum analyze redis
Close: vacuum analyze disk
Close: vacuum analyze kernel
Close: vacuum analyze mem
Close: vacuum analyze net
Close: vacuum analyze postgresl

### DIFF
--- a/cmd/tsbs_load_timescaledb/main.go
+++ b/cmd/tsbs_load_timescaledb/main.go
@@ -53,6 +53,7 @@ func initProgramOptions() (*timescaledb.LoadingOptions, load.BenchmarkRunner, *l
 	opts.PartitionIndex = viper.GetBool("partition-index")
 	opts.FieldIndex = viper.GetString("field-index")
 	opts.FieldIndexCount = viper.GetInt("field-index-count")
+	opts.Analyze = viper.GetBool("analyze")
 
 	opts.ProfileFile = viper.GetString("write-profile")
 	opts.ReplicationStatsFile = viper.GetString("write-replication-stats")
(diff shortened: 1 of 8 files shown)


mdcallag:
Hello reviewer, I am a golang noob. The code runs with and without setting the option. However, I don't know which tests I should run other than doing "go test" in a few directories.

jonatas:
@mdcallag thanks for the contribution. I'll close this PR as you already shared that you're no longer interested in contributing here.

This option looks very interesting but I'm not able to finalize it soon, so let's close it for now and we can reuse it in the future if we go in this direction.

PR #152, opened 2021-01-29
Merging original multi-node branch to allow new SIMULATOR mode

ryanbooz:
In working to ingest and benchmark a large dataset (~200 billion metrics), I merged the original `multi-node` branch into some recent changes change off of `master` which allowed running the `SIMULATOR` into a multi-node setup. The goal would be to then merge this branch from @blagojts into `master`.

I think there should be some discussion about the parameter flag names, or at least how they act together for single vs. multi-node. It's not just a one parameter switch... so we need to just be more clear.

Caveat - I'm not a Golang dev. I understand the basics enough to get the changes into the right place, but I'm sure we need to discuss more tests (😱).  I mostly just didn't want this to be on my laptop only.

### DIFF
--- a/cmd/tsbs_load_timescaledb/main.go
+++ b/cmd/tsbs_load_timescaledb/main.go
@@ -5,7 +5,9 @@ package main
 
 import (
 	"fmt"
+	"log"
 	"sync"
+	"time"
 
 	"github.com/blagojts/viper"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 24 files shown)


pkg/targets/timescaledb/process.go
```diff
@@ -219,17 +222,34 @@ func (p *processor) processCSI(hypertable string, rows []*insertData) uint64 {
...
+	} else {
+		tx := MustBegin(p._db)
+		if p.opts.UseCopy {
+			stmt, err = tx.Prepare(pq.CopyIn(hypertable, cols...))
+			for _, r := range dataRows {
+				stmt.Exec(r...)
+			}
+			_, err = stmt.Exec()
+		} else {
+			stmtString := genBatchInsertStmt(hypertable, cols, len(dataRows))
```
blagojts, line 247: do we still need this now that multinode supports copy?

blagojts (approved):
Looks good, just left some minor questions and one suggestion. 
No need to rereview if changes are applied

niksajakovljevic:
Hey @ryanbooz, thanks for the PR! However we need to rebase these changes and top of current master. Since CI tests are failing maybe rebasing will solve that problem as well. Try running `go test -v -race` locally to make sure that tests are passing.

pkg/targets/timescaledb/process.go
ryanbooz: Good question. I actually wasn't sure about this either, especially after it bit me the first time I tried to run it and didn't have the config setup correctly.

@erimatnor or @niksajakovljevic - any thoughts?

ryanbooz (commented):
Heh... I was going to leave that for the next branch to deal with.  😂

Happy to try.

cmd/tsbs_load_timescaledb/main.go
```diff
@@ -15,12 +17,100 @@ import (
...
+// allows for testing
+var fatal = log.Fatalf
+
 // Parse args:
-func initProgramOptions() (*timescaledb.LoadingOptions, load.BenchmarkRunner, *load.BenchmarkRunnerConfig) {
-	target := timescaledb.NewTarget()
-	loaderConf := load.BenchmarkRunnerConfig{}
-	loaderConf.AddToFlagSet(pflag.CommandLine)
-	target.TargetSpecificFlags("", pflag.CommandLine)
+func init() {
```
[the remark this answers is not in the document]
ryanbooz: You mean in the `program_options.go` file for TimescaleDB? (pkg/targets/timescaledb/program_options.go)

pkg/targets/timescaledb/process.go
erimatnor: If I am not mistaken, I think TSBS at some point switched from COPY to INSERTs (this wasn't multi-node related). I think we just wanted to retain the option to benchmark also the INSERT path. Some people cannot use COPY because it doesn't support upserts, etc.

blagojts:
Fixed the failing test and merged locally in the target branch. Closing

PR #156, opened 2021-03-17
Changing mongo driver

Hadrien-Cornier:
changing mongo driver to the official mongo driver instead of mgo which is no longer being maintained and does not work with mongo 4.4


Hadrien-Cornier:
I changed the tsbs_load_mongo code and the tsbs_run_queries. 
If you want to load data to mongo, you cannot yet use the on the fly tsbs_load mechanism, but you can use tsbs_load_mongo with pre-generated data.

cmd/tsbs_load_mongo/creator.go
```diff
@@ -1,29 +1,46 @@
...
 	if err != nil {
+		log.Println("DID NOT MANAGE TO CONNECT")
 		log.Fatal(err)
+	} else {
+		err = d.client.Ping(d.ctx, readpref.Primary())
+		if err != nil {
+			log.Println("DID NOT MANAGE TO CONNECT")
+			log.Fatal(err)
+		} else {
+			log.Println("MANAGED TO CONNECT")
```
jonatas, line 36: Can you please clean up the debugging messages? I also see it uses ping to confirm the server it responding, which I think is not needed for the connection.
```suggestion
	d.ctx, d.cancel = context.WithTimeout(context.Background(), writeTimeout)
	d.client, err = mongo.Connect(d.ctx, options.Client().ApplyURI(daemonURL))
	if err != nil {
		log.Println("Can't establish connection with", daemonURL)
		log.Fatal(err)
```

cmd/tsbs_load_mongo/creator.go
```diff
@@ -36,74 +53,68 @@ func (d *dbCreator) DBExists(dbName string) bool {
...
-	cmd = append(cmd, bson.DocElem{
-		Name: "storageEngine", Value: map[string]interface{}{
-			"wiredTiger": map[string]interface{}{
-				"configString": "block_compressor=snappy",
-			},
-		},
-	})
-
-	err := d.session.DB(dbName).Run(cmd, nil)
+	//log.Println("tsbs_load_mongo/creator/CreateDB")
```
jonatas, line 74:
```suggestion
```

cmd/tsbs_load_mongo/creator.go
```diff
@@ -36,74 +53,68 @@ func (d *dbCreator) DBExists(dbName string) bool {
...
+			panic(err)
 		}
 	}
-
 	return nil
 }
 
 func (d *dbCreator) Close() {
-	d.session.Close()
+	//closing the database connection here
```
jonatas, line 118: Can we fix the error and wait for the bulk finish and then close the connection? or force close in case that is what the user wants to do?

cmd/tsbs_load_mongo/document_per_loader.go
```diff
@@ -1,13 +1,14 @@
 package main
 
 import (
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 	"log"
 	"sync"
-
-	"github.com/globalsign/mgo"
+	//"github.com/globalsign/mgo"
```
jonatas, line 8: Please, remove unused packages:
```suggestion
```

cmd/tsbs_load_mongo/document_per_loader.go
```diff
@@ -85,9 +85,10 @@ func (p *naiveProcessor) ProcessBatch(b targets.Batch, doLoad bool) (uint64, uin
...
 
 	if doLoad {
-		bulk := p.collection.Bulk()
-		bulk.Insert(p.pvs...)
-		_, err := bulk.Run()
+		//bulk := p.collection.Bulk()
+		opts := options.InsertMany().SetOrdered(false)
+		//bulk.Insert(p.pvs...)
+		_, err := p.collection.InsertMany(p.dbc.ctx, p.pvs, opts)
 		if err != nil {
```
jonatas, line 92:
```suggestion
		opts := options.InsertMany().SetOrdered(false)
		_, err := p.collection.InsertMany(p.dbc.ctx, p.pvs, opts)
		if err != nil {
```

cmd/tsbs_load_mongo/creator.go
Hadrien-Cornier: This is tricky, I'm going to look into it when I have more time

jonatas (changes requested):
Thanks for the update! I think we need some clean-ups to proceed, let me know if you need any help.

cmd/tsbs_run_queries_mongo/main.go
```diff
@@ -5,46 +5,53 @@
...
+	"context"
 	"encoding/gob"
 	"fmt"
-	"log"
-	"time"
-
 	"github.com/blagojts/viper"
-	"github.com/globalsign/mgo"
-	"github.com/globalsign/mgo/bson"
+	mgobson "github.com/globalsign/mgo/bson"
```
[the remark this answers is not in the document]
jonatas: oush, I was learning this is just a way to rename the package locally 🤦

PR #168, opened 2021-06-18
Add multinode to master

ryanbooz:
This was attempted quite a while ago and in changes before @atanasovskib left we apparently missed finishing that original attempt. 

This PR merges in the `multinode` branch and reworks a few of the parameters that determine whether or not we create a distributed hypertable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,3 @@ jobs:
       install: skip
       script:
         - GO111MODULE=on go test -v -race -coverprofile=coverage.txt -covermode=atomic ./...
-      after_success:
-        - bash <(curl -s https://codecov.io/bash)
--- a/cmd/tsbs_generate_queries/databases/cassandra/devops.go
+++ b/cmd/tsbs_generate_queries/databases/cassandra/devops.go
@@ -111,8 +111,8 @@ func (d *Devops) GroupByTimeAndPrimaryTag(qi query.Query, numMetrics int) {
 // FROM cpu WHERE (hostname = '$HOSTNAME_1' OR ... OR hostname = '$HOSTNAME_N')
 // AND time >= '$HOUR_START' AND time < '$HOUR_END'
 // GROUP BY hour ORDER BY hour
-func (d *Devops) MaxAllCPU(qi query.Query, nHosts int) {
-	interval := d.Interval.MustRandWindow(devops.MaxAllDuration)
+func (d *Devops) MaxAllCPU(qi query.Query, nHosts int, duration time.Duration) {
+	interval := d.Interval.MustRandWindow(duration)
 
 	tagSet := d.getHostWhere(nHosts)
 
--- a/cmd/tsbs_generate_queries/databases/clickhouse/devops.go
+++ b/cmd/tsbs_generate_queries/databases/clickhouse/devops.go
@@ -85,8 +85,8 @@ const clickhouseTimeStringFormat = "2006-01-02 15:04:05"
 // Resultsets:
 // cpu-max-all-1
 // cpu-max-all-8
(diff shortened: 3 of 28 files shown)


ryanbooz:
If you get a chance early in the week @antekresic, I'd love some help understanding how to set some defaults through Viper. I've looked at a bunch of their docs and GitHub issues detailing how it should work, but all of my attempts don't seem to actually set default values to be used if the parameter isn't contained in the YAML file.  Talking mostly about `cmd -> tsbs_load_timescaledb -> main.go`

jonatas (commented):
Very good! I couldn't test in a cluster as the replication factor still not available in the command line.

Here are the steps I did to try to run it:

1. checkout the branch
2. make all # refresh all executables

When I run with the argument I receive:

```
tsbs_load_timescaledb --help                                                         18:07:22
Usage of tsbs_load_timescaledb:
      --admin-db-name user               Database to connect to in order to create additional benchmark databases.
                                         By default this is the same as the user (i.e., `postgres` if neither is set),
                                         but sometimes a user does not have its own database. (default "postgres")
      --batch-size uint                  Number of items to batch together in a single insert (default 10000)
      --chunk-time duration              Duration that each chunk should represent, e.g., 12h (default 12h0m0s)
      --create-metrics-table             Drops existing and creates new metrics table. Can be used for both regular and hypertable (default true)
      --db-name string                   Name of database (default "benchmark")
      --do-abort-on-exist                Whether to abort if a database with the given name already exists.
      --do-create-db                     Whether to create the database. Disable on all but one client if running on a multi client setup. (default true)
      --do-load                          Whether to write data. Set this flag to false to check input read speed. (default true)
      --field-index string               index types for tags (comma delimited) (default "VALUE-TIME")
      --field-index-count int            Number of indexed fields (-1 for all)
      --file string                      File name to read data from
      --force-text-format                Send/receive data in text format
      --hash-workers                     Whether to consistently hash insert data to the same workers (i.e., the data for a particular host always goes to the same worker)
      --host string                      Hostname of TimescaleDB (PostgreSQL) instance (default "localhost")
      --in-table-partition-tag           Whether the partition key (e.g. hostname) should also be in the metrics hypertable
      --insert-intervals string          Time to wait between each insert, default '' => all workers insert ASAP. '1,2' = worker 1 waits 1s between inserts, worker 2 and others wait 2s
      --limit uint                       Number of items to insert (0 = all of them).
      --log-batches                      Whether to time individual batches.
      --partition-index                  Whether to build an index on the partition key (default true)
      --partitions int                   Number of partitions
      --pass string                      Password for user connecting to PostgreSQL (leave blank if not password protected)
      --port string                      Which port to connect to on the database host (default "5432")
      --postgres string                  PostgreSQL connection string (default "sslmode=disable")
      --reporting-period duration        Period to report write stats (default 10s)
      --seed int                         PRNG seed (default: 0, which uses the current timestamp)
      --time-index                       Whether to build an index on the time dimension (default true)
      --time-partition-index             Whether to build an index on the time dimension, compounded with partition
      --use-hypertable                   Whether to make the table a hypertable. Set this flag to false to check input write speed against regular PostgreSQL. (default true)
      --use-jsonb-tags                   Whether tags should be stored as JSONB (instead of a separate table with schema)
      --user string                      User to connect to PostgreSQL as (default "postgres")
      --workers uint                     Number of parallel clients inserting (default 1)
      --write-profile string             File to output CPU/memory profile to
      --write-replication-stats string   File to output replication stats to
pflag: help requested
```

And the `replication-factor` param is not there. @ryanbooz can you share the steps I can take to test it properly?

ryanbooz:
Great catch! I neglected to put them in the file that sets up command-line input. Added the required flags.

jonatas (approved):

PR #169, opened 2021-06-22
Parametrize full mini test

jonatas:
Work in progress. Our idea is to have a single command to run and share variables through different settings. As a first step we're:

1. Renaming from full_cycle_minitest to full_cycle
2. Creating a set of variables that can be used to configure
3. Loop over the scenarios to make our life easier.


scripts/full_cycle/full_cycle_timescaledb.sh
```diff
@@ -0,0 +1,89 @@
...
+
+
+CONTAINER_NAME=timescaledb_benchmark
+# setup pg with password and expose the default port.
+sudo docker run -d --name $CONTAINER_NAME -p 5432:$PORT \
+  -e POSTGRES_PASSWORD=$PASSWORD \
+  timescale/timescaledb:latest-pg12
+
+docker start $CONTAINER_NAME
+sleep 2
```
ryanbooz, line 23: Maybe we wrap this in a check in case we're not running this locally?

PR #175, opened 2021-08-13
victoriametrics: follow-up after https://github.com/timescale/tsbs/pull/168

hagen1778:
The interface change in https://github.com/timescale/tsbs/pull/168
caused unexpected panics for VictoriaMetrics mode.
This commit updates the interface implementation.

It also updates docs and load script with adding more details.


jonatas (approved):
Thanks for the fix @hagen1778! Looks good to me!

PR #181, opened 2021-11-04
Add QuestDB to makefile

puzpuzpuz:
`tsbs_load_questdb` and `tsbs_run_queries_questdb` binaries were not generated when building tsbs with `make`


jonatas (approved):
Thank you, @puzpuzpuz!

PR #195, opened 2022-03-30
generate consistent query conditions using the same random seed

huskar-t:


ryanbooz:
@huskar-t - can you please explain the intent of this PR. It's unclear what the impact is by moving the call to this function throughout the project. Thanks!

huskar-t:
@ryanbooz 
since both the generation conditions and time ranges use random functions, different db targets obtain time ranges and generate conditions in different orders in the same query scenario, resulting in different query statements for the same random seed.

huskar-t:
@RobAtticus any suggestion？

PR #233, opened 2023-07-04
Fix panic in timescale target with `iot` use case

arajkumar:


arajkumar:
@antekresic Could you please review this?

antekresic (approved):

PR #234, opened 2023-08-15
docs: add `Troubleshooting` and `Expected numbers` for VictoriaMetrics

hagen1778:
Update docs for VictoriaMetrics to point out possible bottlenecks and provide expected numbers.


hagen1778:
Hello! The Travis check fails without good reason... Should I do something with it?

hagen1778:
Hello! Any updates?

PR #240, opened 2023-11-02
Initial implementation of OpenMetrics text format

qoega:
Added support of OpenMetrics(Prometheus compatible) text format.
Spec: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#text-format

Important details:
- Different fieldValues are flattened to separate rows.
- Empty value is not supported, so nil fieldValues are not exported.

As OpenMetrics is a common spec I think load should not test only one system, but it is possible to test OpenMetrics data load to different systems that support this ingestion format. So it kind of adds another dimension to harness.


qoega:
Can we sign Corporate CLA instead?

PR #254, opened 2024-06-06
Fixed influx data format parsing

Koheiru:
Influx line parsing improved because influx line protocol allows to use space and comma symbols in tag name, tag value, field name and field value. In this case such symbols should be escaped or quoted.


Koheiru:
The reason why this improvement needs is that I also have additional changes about new data set that contains complex string tags and fields.

Issue #248, opened 2024-03-12
tsbs_load load timescaledb error


ruoshui9527:
tsbs_generate_data --use-case="iot" --seed=123 --scale=4000 \
    --timestamp-start="2024-03-10T00:00:00Z" \
    --timestamp-end="2024-03-12T00:00:00Z" \
    --log-interval="10s" --format="timescaledb" \
    | gzip > /tmp/timescaledb-data.gz

## use tsbs_load:
./tsbs_load config --target=timescaledb --data-source=FILE

vi After modifying the relevant content, execute
./tsbs_load load timescaledb --config=./config.yaml
Using config file: ./config.yaml
2024/03/12 11:50:22 input header in wrong format. got '%!s(uint8=31)', expected 'tags'

##use tsbs_load_timescaledb:
cat /tmp/timescaledb-data.gz | gunzip | ./tsbs_load_timescaledb \
> --postgres="sslmode=disable" --host="192.168.1.10" --port=5433 --pass="111111" \
> --user="postgres" --admin-db-name=tsbs --workers=8  \
> --in-table-partition-tag=true --chunk-time=8h --write-profile= \
> --field-index-count=1 --do-create-db=true --force-text-format=false \
> --do-abort-on-exist=false --create-metrics-table=true
could not execute sql: SELECT create_hypertable('readings'::regclass, 'time'::name, replication_factor => NULL, chunk_time_interval => 28800000000, create_defe_hypertable(regclass, name, replication_factor => unknown, chunk_time_interval => bigint, create_default_indexes => boolean) does not exist (SQLSTATE 42883)

goroutine 1 [running]:
github.com/timescale/tsbs/pkg/targets/timescaledb.MustExec(0x8d8971?, {0xc0000e6000, 0x9b}, {0x0?, 0x4?, 0x8?})
        /home/ps/tsbs-master/pkg/targets/timescaledb/creator.go:296 +0xd0
github.com/timescale/tsbs/pkg/targets/timescaledb.(*dbCreator).createTableAndIndexes(0xc00008a0f0, 0xc000128dd0, {0xc000178140, 0x8}, {0xc00016a280, 0x8, 0x8}
        /home/ps/tsbs-master/pkg/targets/timescaledb/creator.go:224 +0x6ff
github.com/timescale/tsbs/pkg/targets/timescaledb.(*dbCreator).PostCreateDB(0xc00008a0f0, {0x8bc0b4?, 0x9?})
[...]

PR #255, opened 2024-06-06
Fixed timescaledb loader initialization

Koheiru:
At some point timescaledb's function `create_hypertable` became more strict and it starts to fail if 'replication_factor => NULL' part exists. In such case `tsbs_load` failed with error:
```
could not execute sql: SELECT create_hypertable('disk'::regclass, 'time'::name, replication_factor => NULL, chunk_time_interval => 43200000000, create_default_indexes=>FALSE)panic: ERROR: function create_hypertable(regclass, name, replication_factor => unknown, chunk_time_interval => bigint, create_default_indexes => boolean) does not exist (SQLSTATE 42883)
```

Other users also report about this issue: #248 

To fix this issue timescaledb table initialization method was adjusted.

### DIFF
--- a/pkg/targets/timescaledb/creator.go
+++ b/pkg/targets/timescaledb/creator.go
@@ -195,10 +195,11 @@ func (d *dbCreator) createTableAndIndexes(dbBench *sql.DB, tableName string, fie
 	}
 
 	if d.opts.UseHypertable {
-		var creationCommand string = "create_hypertable"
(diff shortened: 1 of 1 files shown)


tomchon:
has anyone merged this fix into the master branch? @Koheiru

basheer-taplane:
Why is this not merged yet? I am having the same issue.

orcarex:
faced the same issue

Issue #275, opened 2025-05-18
tsbs_load_clickhouse: Import 100 devices but the tags table has 10,000 rows.


cocobond:
## Test steps:
1.  
./tsbs_generate_data --use-case="cpu-only" \
  --format="clickhouse" \
  --log-interval="10s" \
  --scale=100 \
  --timestamp-start="2024-01-01T00:00:00Z" \
  --timestamp-end="2024-02-01T00:00:00Z" \
  --file="ck_cpu_data_100_devices.csv"
> 生成100devices的预备导入数据
![Image](https://github.com/user-attachments/assets/83159e15-4053-45eb-8067-e99650cbc06d)

2. 
After the import is complete, the tags table has 10,000 device IDs, of which the first 9,900 rows are imported incorrectly, and only the last 100 are valid device IDs (referring to the corresponding tags_id in the cpu table)
![Image](https://github.com/user-attachments/assets/d66680d7-e523-4499-a492-632b635b1c5f)

PR #276, opened 2025-05-18
Fix bug that load_clickhouse resulted in wrong number of devices in t…

cocobond:
## After FIX: 
image

> Since only the first batch will follow this code logic, and subsequent batches will follow another function, it will not affect other scenarios. This simply fixes this bug.

This bug has been fixed. 🥳 
I guess it was because clickhouse support imitated the writing of Timescale, but in the table creation statement of Timescale, hostname unique and tags_id nextval were not deduplicated and had no impact...


cocobond:
Fixes https://github.com/timescale/tsbs/issues/275 🙌

